PR #3, opened 2018-10-08
Replace MoveTowards with Lerp

BenNaccarato:
#### Description
Replace the MoveTowards lerp with a real lerp - use Duration, rather than speed to move to a new value.

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/Animation/FpsAnimator.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/Animation/FpsAnimator.cs
@@ -22,8 +22,19 @@ namespace Fps
             public string Name;
             public float OldValue;
             public float CurrentValue;
-            public float TargetValue;
             public float LerpTimeElapsed;
+            private float targetValue;
+
+            public float TargetValue
+            {
+                get => targetValue;
+                set
+                {
+                    OldValue = CurrentValue;
+                    targetValue = value;
+                    LerpTimeElapsed = 0;
+                }
+            }
         }
 
         private class AnimationBoolParameter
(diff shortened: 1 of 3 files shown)


workers/unity/Assets/Fps/Scripts/GameLogic/Animation/FpsAnimator.cs
```diff
@@ -195,33 +197,59 @@ private static float CalculateAngle(Transform from, Vector3 direction)
...
-            LerpFloatParameter(xParameter, animationSettings.InterpolateMovementSpeed);
-            LerpFloatParameter(zParameter, animationSettings.InterpolateMovementSpeed);
-            LerpFloatParameter(movementParameter, animationSettings.InterpolateMovementSpeed);
+            LerpFloatParameter(pitchParameter, animationSettings.InterpolatePitchDuration);
+            LerpFloatParameter(xParameter, animationSettings.InterpolateMovementDuration);
+            LerpFloatParameter(zParameter, animationSettings.InterpolateMovementDuration);
+            LerpFloatParameter(movementParameter, animationSettings.InterpolateMovementDuration);
+        }
+
+        private static void SetFloatParameter(AnimationFloatParameter parameter, float newValue)
```
jamiebrynes7: Make this a setter on the `AnimationFloatParameter.TargetValue` property so we can get rid of this function

BenNaccarato: Ah yes, will do

peterimprobable:
Definite improvement. Looks good in local tests.
Lerp > MoveTowards for animations!

PR #4, opened 2018-10-09
Add Setup Scripts (for real this time)

jamiebrynes7:
#### Description
Added switch for selecting SSH or HTTPS for clone at the request of @jessicafalk 
#### Tests
Tested both scripts on Powershell and Git Bash
#### Documentation
Will be!
#### Primary reviewers


scripts/powershell/setup.ps1
```diff
@@ -0,0 +1,87 @@
...
+if ($CloneUri -eq -1)
+{
+    Write-Host "Stopping the setup process."
+    exit 0
+}
+
+Write-Warning @"
+
+This script will create the following directory: $($TargetDirectory)
+If such a directory already exists it will be deleted.
```
jared-improbable: Add some "*** If such a directory already exists it will be deleted. ***" or write the line in a different colour.

scripts/powershell/setup.ps1
```diff
@@ -0,0 +1,87 @@
...
+$PinnedVersion = Get-Content ($PSScriptRoot + "/../../gdk.pinned")
+
+Write-Host "Checking out pinned version: $($PinnedVersion)" -ForegroundColor Yellow
+
+Push-Location -Path "$TargetDirectory"
+    & git checkout "$PinnedVersion"
+    
+    if ($LastExitCode -ne 0)
+    {
+        exit 1
```
jared-improbable: Will exiting early result in your owning shell being left in `$TargetDirectory`?

scripts/shell/setup.sh
```diff
@@ -0,0 +1,75 @@
...
+echo "Select a cloning method."
+echo ""
+options=("HTTPS" "SSH" "Quit")
+select opt in "${options[@]}"
+do
+    case $opt in
+        "HTTPS")
+            CLONE_URI="https://github.com/spatialos/gdk-for-unity.git"
+            break
+            ;;
```
jared-improbable: What a well-designed "language"

scripts/powershell/setup.ps1
```diff
@@ -0,0 +1,87 @@
+$ErrorActionPreference = "Stop"
+
+function Select-Clone-Method
+{
+    Write-Host @"
+Select a cloning method.
+
+Press 1 for HTTPS
```
firtina-improbable: optional: may help to say what the URI would look like here, because some people may not know what "ssh" means in this context

scripts/shell/setup.sh
jamiebrynes7: If only we could write these in a sane scripting language

scripts/powershell/setup.ps1
jamiebrynes7: Ah yes I think so- will fix

scripts/shell/setup.sh
firtina-improbable: request changes: pls use javascript

jamiebrynes7: I said **a sane scripting language** 😡

PR #14, opened 2018-10-09
remove all bullets if player disabled/removed

BenNaccarato:
#### Description
A tentative bug fix for some linerenderers remaining in the world. (Coroutines are stopped, so if a player is removed, their existing bullets would not be sent back to the pool)
This hasn't been tested in a deployment yet.


peterimprobable:
How so?
It would result in bullets vanishing if the user who fired them moved out of checkout range, but this is very much an edge case.
A potentially better way would be to move all logic into the Shot/LineRenderer GO itself, as this makes them fire/forget, but we would potentially lose some of the barrel snapping we currently have.

Given the trade off, I think this is probably the best approach.

jamiebrynes7:
That's fair enough 👍

PR #15, opened 2018-10-09
Change the forced rotation a player receives on respawn, from a comma…

peterimprobable:
Sami raised the issue that we are having performance issues with commands.
It is therefore desirable to remove command usage where possible.
It also appears the command wasn't even being used, and was throwing erros logs.
Therefore an event, with actual listen functionality, seems like a better approach.

To test:
Get shot while pointing in a specifically odd angle, respawn, check that you are now pointing at the angle that the respawn point wanted you to point.
Also check a fake-player heavy deployment, for lack of rotation command error lgos.

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/FpsDriver.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/FpsDriver.cs
@@ -60,7 +60,7 @@ namespace Fps
         {
             Cursor.lockState = CursorLockMode.Locked;
             Cursor.visible = false;
-            serverMovement.OnForcedRotation += OnForcedRotaiton;
+            serverMovement.OnForcedRotation += OnForcedRotation;
         }
 
         private void Update()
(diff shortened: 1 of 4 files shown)


workers/unity/Assets/Fps/Scripts/GameLogic/FpsDriver.cs
```diff
@@ -197,6 +199,12 @@ private void Animations(bool isJumping)
             }
         }
 
+        private void OnForcedRotaiton(RotationUpdate forcedRotaiton)
```
jamiebrynes7: Nit: `Rotaiton` -> `Rotation`

jamiebrynes7: And in the argument

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/ClientMovementDriver.cs
```diff
@@ -148,6 +148,29 @@ private void OnEnable()
             spatialOSComponent = GetComponent<SpatialOSComponent>();
             origin = spatialOSComponent.Worker.Origin;
             server.LatestUpdated += OnServerUpdate;
+            server.OnForcedRotation += OnForcedRotation;
+        }
+
+        private void OnForcedRotation(RotationUpdate forcedRotaiton)
```
jamiebrynes7: `Rotaiton`

PR #20, opened 2018-10-09
Removal of the commented out "Desert" building code in the MapBuilder

peterimprobable:
The map builder code contained a chunk of commented out code for building the "desert" area.

This should either be:
  Removed
  Reenabled
  Parameterised

This PR is the "Remove" option, and also renames the remaining wall to "Surround" due to "Desert" no longer making sense.
However I did leave the desert plane and cube prefabs in the repo, which could arguably be removed.

Do not merge unless we have agreement that this is the desired solution.


workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -257,49 +253,19 @@ private void PlaceEdges(int groundLayers)
...
     {
-        // desertPlane = Resources.Load<GameObject>(DesertPlanePath);
-        // if (desertPlane == null)
-        // {
-        //     Debug.LogError("Desert plane prefab not loaded.");
-        //     return;
-        // }
-
-        desertParentTransform = new GameObject(DesertParentName).transform;
+        desertParentTransform = new GameObject(SurroundParentName).transform;
```
paulbalaji: we should probably rename this variable to `surroundParentTransform` as well

peterimprobable:
Good catch, renamed all remaining references to "Desert" in the script

PR #26, opened 2018-10-10
Feature/loadscreen ui 2.0

amandakys:
#### Description
Changed the loading screen flow to only connect client worker after the player clicks ready to connect. 

NOTE: the respawn/retry UI flow is build based on the fact that the client sends create player requests every frame once the client worker has been connected. To avoid the player going straight from the failed screen into the game, a wait time of 2 seconds in enforced and players must go to the "connecting" screen first before the UI is removed from view. Behind the UI, the game is actually already loaded. This hacky way of displaying UI can be fixed by modifying the way the playerlifecycle sends create player requests.


BenNaccarato:
- So it looks like the connect buttons don't work - (Pressing spacebar does work fine though). I'm guessing something needs to be hooked up again?
- Also, I don't think we have a transition to a 'Failed to connect' - the client seems to be stuck in connecting (If connecting fine to SpatialOS, but failing to spawn a player).

jessicafalk:
closing it as it has become obsolete

PR #29, opened 2018-10-10
Prevent multiple ragdolls being spawned on death

peterimprobable:
Currently multiple death events are being received per death.
Added a boolean flag to prevent multiple ragdolls spawning.

Ideally we would solve the root cause, but in the meantime, this is visually a very nasty bug, so a hotfix will improve the user experience in the short term.

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.ragdolls/Behaviours/RagdollSpawner.cs
+++ b/workers/unity/Packages/com.improbable.gdk.ragdolls/Behaviours/RagdollSpawner.cs
@@ -17,18 +17,18 @@ namespace Improbable.Gdk.Ragdoll
         public RagdollSpawned OnRagdollSpawned;
 
         private ObjectPool<PoolableRagdoll> pool;
-        private bool ragdollSpawned;
+        private bool hasRagdollSpawned;
 
         private void OnEnable()
         {
             health.OnHealthModified += OnHealthModified;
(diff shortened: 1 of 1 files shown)


workers/unity/Packages/com.improbable.gdk.ragdolls/Behaviours/RagdollSpawner.cs
```diff
@@ -17,23 +17,26 @@ public class RagdollSpawner : MonoBehaviour
         public RagdollSpawned OnRagdollSpawned;
 
         private ObjectPool<PoolableRagdoll> pool;
+        private bool ragdollSpawned;
```
jamiebrynes7: Nit: `hasRagdollSpawned`

workers/unity/Packages/com.improbable.gdk.ragdolls/Behaviours/RagdollSpawner.cs
```diff
@@ -17,23 +17,26 @@ public class RagdollSpawner : MonoBehaviour
...
 
         private void OnEnable()
         {
             health.OnHealthModified += OnHealthModified;
+            health.OnRespawn += empty => ragdollSpawned = false;
```
jamiebrynes7: Can you use the `(empty) => ...` syntax here? I feel it makes it clearer?

PR #31, opened 2018-10-10
Remove GroundChecker, replace with methods in CharacterControllerMotor

BenNaccarato:
#### Description
The groundchecker had some duplicated functionality with the ClientMovementDriver, and used an expensive ground checking method so we decided to reformat, removing the GroundChecker entirely, and putting the ground checking functionality in the base CharacterControllerMotor class.


workers/unity/Assets/Fps/Resources/Prefabs/UnityGameLogic/Player.prefab
```diff
@@ -99,6 +87,10 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   teleportDistanceSqrd: 25
+  groundedRadius: 0.05
```
peterimprobable: What's the reason for the much smaller grounded radius here?

BenNaccarato: No, that's my bad - hadn't noticed the ServerMovementDriver - tbh, it doesn't require a grounded check. I could experiment with hiding it, or maybe I should change the inheritance of the classes such that it doesn't include the ground checking code?

workers/unity/Assets/Fps/Resources/Prefabs/SimulatedPlayer/SimulatedPlayer.prefab
```diff
@@ -186,6 +170,10 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   teleportDistanceSqrd: 25
+  groundedRadius: 0.2
```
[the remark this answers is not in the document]
BenNaccarato: Yup, that value seems to work alright - The character controller has a radius of 0.25, this means that you remain grounded for most of your radius, and only slip off in that outer 0.05

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/ProxyMovementDriver.cs
```diff
@@ -87,5 +86,11 @@ protected override void Update()
                 hasRotationLeft = false;
             }
         }
+
+        public bool GetGrounded()
```
peterimprobable: Is there a way we can just call CheckGrounded after each movement change is applied, and no need for this method as we can just access IsGrounded? (To avoid doing the check every frame on things that aren't even moving)

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/CharacterControllerMotor.cs
```diff
@@ -109,17 +117,22 @@ public virtual void MoveWithExtensions(Vector3 toMove)
...
         }
 
-        protected bool CheckOverrideInAir()
+        protected void CheckGrounded()
+        {
+            IsGrounded = Physics.OverlapSphereNonAlloc(transform.position, groundedRadius, groundedOverlapSphereArray,
+                groundLayerMask) > 0;
+        }
+
+        protected void CheckOverrideInAir()
```
peterimprobable: Confused by the name of this method. What's the meaning of "Override" in this context? Was this an already existing method name?

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/ProxyMovementDriver.cs
BenNaccarato: Yeah, that can be done, but may involve a number of unneeded calculations, if ground checking isn't needed, or multiple `move`s are called in a frame. But I agree, it is neater

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/CharacterControllerMotor.cs
BenNaccarato: This was, yeah. So the extensions can force the motor to be considered as not grounded, despite the ground checker detecting it and passing. I can rename this if something would be preferable?
(CheckExtensionsForOverrides?)

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/ProxyMovementDriver.cs
peterimprobable: Is the grounded state currently being called every frame by the animator? If so, the change shoudl lead to less calculations, if anything.

workers/unity/Packages/com.improbable.gdk.movement/Behaviours/CharacterControllerMotor.cs
peterimprobable: Hmm, yeah, that makes a bit more sense I guess.

PR #32, opened 2018-10-10
MacOS Setup Script Fix

jamiebrynes7:
#### Description
If a user has MacOS and has not installed `coreutils` through `brew` they will not have `realpath`. This PR introduces a workaround for this - but at the cost of it no longer follows symlinks. This is probably okay - and the user has to provide confirmation to the path given to them.
#### Tests
Tested on MacOS and Git Bash
#### Documentation
N/A
#### Primary reviewers
@jared-improbable @jessicafalk

### DIFF
--- a/scripts/shell/setup.sh
+++ b/scripts/shell/setup.sh
@@ -19,8 +19,16 @@ echo ""
 echo "Welcome to the SpatialOS GDK for Unity FPS Starter Project setup script." 
 echo ""
 
-readonly RAW_DIR="$(CDPATH= cd -- "$(dirname -- $0)" && pwd)/../../../gdk-for-unity"
-readonly TARGET_DIRECTORY="$(realpath "${RAW_DIR}")"
+readonly RAW_DIR="$(dirname "${0}")/../../../gdk-for-unity"
+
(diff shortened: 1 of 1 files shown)


jamiebrynes7:
Yeah I'm not going to merge before testing more thoroughly 👍

scripts/shell/setup.sh
```diff
@@ -19,7 +19,7 @@ echo ""
 echo "Welcome to the SpatialOS GDK for Unity FPS Starter Project setup script." 
 echo ""
 
-readonly RAW_DIR="$(dirname "${0}")/../../../gdk-for-unity"
+readonly RAW_DIR="$(CDPATH= cd -- "$(dirname -- $0)" && pwd)/../../../gdk-for-unity"
 readonly TARGET_DIRECTORY="$(realpath "${RAW_DIR}")"
```
mvriel: Still getting the error due to the realpath on this line

jamiebrynes7: Wow I'm so tired I replaced the wrong line 😱

jamiebrynes7:
Validated on Mac TC agents

PR #35, opened 2018-10-11
Bug fix - aiming up

BenNaccarato:
Note: The gun bullet snapping is still quite wonky - The fact that the fudge factor is recalculated every update probably doesn't help. May be worth experimenting with the rotation speeds and the assorted thresholds to see if it helps (Build out clients seem to bullet snap fine - may be that the gun's pitch changes too suddenly? Possibly after shooting it snaps to something?)
(Used the transform rotation of the object (which is constrained to 0 pitch), rather than the actual pitch of the gun)


jessicafalk:
closing it as it has become obsolete

PR #42, opened 2018-10-15
[Fix] players stuck after respawn

mattyoung-improbable:
#### Description
1. Delay updating spatial position when respawning player to the next frame, this ensures that the other component updates all get updated before the potential worker transition.
1. Refactor VisibiltyAndCollision to listen to health updates and not respawn/healthmodified events since these may not arrive after worker transition.
#### Tests
Tested in local deployment, and in cloud deployment. Did not get stuck, or see simulated players floating in the air

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
@@ -1,5 +1,4 @@
 using System.Collections.Generic;
-using Improbable.Common;
 using Improbable.Gdk.GameObjectRepresentation;
 using Improbable.Gdk.Health;
 using UnityEngine;
@@ -10,7 +9,7 @@ namespace Fps
     {
         [Require] private HealthComponent.Requirable.Reader health;
 
-        private bool isDead = false;
+        private bool isVisible = true;
 
         private CharacterController characterController;
         [SerializeField] private List<Renderer> renderersToIgnore = new List<Renderer>();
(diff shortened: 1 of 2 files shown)


workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
```diff
@@ -22,60 +21,36 @@ private void Awake()
...
-            // Enable/Disable the character controller (to turn on/off collision)
             if (characterController)
             {
-                characterController.enabled = enabled;
+                characterController.enabled = visible;
             }
 
-            // Enable/Disable the renderers to turn on/off the model's visibility.
-            foreach (var renderer in GetComponentsInChildren<Renderer>())
+            foreach (var r in GetComponentsInChildren<Renderer>())
```
jamiebrynes7: `r` -> `renderer` or `childRenderer`

workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
```diff
@@ -22,60 +21,36 @@ private void Awake()
 
         private void OnEnable()
         {
-            health.OnRespawn += OnRespawn;
-            health.OnHealthModified += OnHealthModified;
-            ApplyState();
-            SetIsDeadState(health.Data.Health <= 0);
+            health.HealthUpdated += HealthUpdated;
+            SetIsVisible(health.Data.Health > 0);
```
jamiebrynes7: This doesn't need to pass in a parameter as the only two times its called is with `health.Data.Health > 0` as the argument - can just move that into the function

jamiebrynes7: And then rename to `UpdateVisibility`

workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
mattyoung-improbable: done.

workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
mattyoung-improbable: Done

workers/unity/Assets/Fps/Scripts/GameLogic/VisibilityAndCollision.cs
```diff
@@ -22,11 +22,23 @@ private void Awake()
...
+        {
+            UpdateVisibility();
+        }
+
+        private void UpdateVisibility()
         {
+            if (health == null)
+            {
+                return;
+            }
```
jamiebrynes7: This null check shouldn't ever be true given where this is called - but better safe than sorry I guess 😛

PR #68, opened 2018-11-27
Dynamically create ObjectPooler instance

paulbalaji:
**Contributions**: We are not currently taking public contributions - see our [contributions](https://docs.improbable.io/unity/alpha/contributing) policy. However, we are accepting issues and we do want your [feedback](https://github.com/spatialos/gdk-for-unity#give-us-feedback).

-------

#### Description
* remove `ObjectPooling` gameobject from all scenes
* create `ObjectPool` gameobject with the `ObjectPooler` monobehaviour at runtime

Simply removing the `ObjectPooler` instance from scenes results in numerous errors relating to the missing instance, so the `ObjectPooler` instance now gets created at runtime if it's found to be null when calling `GetOrCreateObjectPool<T>`.

#### Tests
- [x] tested locally in development scene
- [x] test with a cloud deployment

#### Documentation
n/a

#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Fps/Scenes/FPS-Client.unity
+++ b/workers/unity/Assets/Fps/Scenes/FPS-Client.unity
@@ -300,47 +300,7 @@ Transform:
   m_LocalScale: {x: 1, y: 1, z: 1}
   m_Children: []
   m_Father: {fileID: 0}
-  m_RootOrder: 6
-  m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
---- !u!1 &229366279
-GameObject:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 0}
-  serializedVersion: 6
-  m_Component:
-  - component: {fileID: 229366281}
-  - component: {fileID: 229366280}
-  m_Layer: 0
-  m_Name: ObjectPooling
-  m_TagString: Untagged
-  m_Icon: {fileID: 0}
-  m_NavMeshLayer: 0
-  m_StaticEditorFlags: 0
-  m_IsActive: 1
---- !u!114 &229366280
-MonoBehaviour:
-  m_ObjectHideFlags: 0
-  m_CorrespondingSourceObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 0}
-  m_GameObject: {fileID: 229366279}
-  m_Enabled: 1
-  m_EditorHideFlags: 0
-  m_Script: {fileID: 11500000, guid: e7ba79fce617e744387e4dca3103886d, type: 3}
-  m_Name: 
-  m_EditorClassIdentifier: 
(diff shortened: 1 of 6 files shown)


workers/unity/Assets/Fps/Scenes/FPS-Client.unity
```diff
@@ -1159,15 +1119,15 @@ MonoBehaviour:
   edgeLeakFixup: 1
   edgeToleranceNormal: 1
   edgeTolerance: 1
-  shadowCascadeCount: 2
+  shadowCascadeCount: 3
```
zeroZshadow: I really wonder why these always change

workers/unity/Packages/com.improbable.gdk.objectpooling/ObjectPooler.cs
```diff
@@ -18,6 +18,11 @@ private void Awake()
         public static ObjectPool<T> GetOrCreateObjectPool<T>(GameObject prefab, int prePoolCount = 0)
             where T : IPoolableObject
         {
+            if (Instance == null)
+            {
+                new GameObject("ObjectPool").AddComponent<ObjectPooler>();
```
zeroZshadow: Please add a comment that the creation of the Component sets `Instance`

workers/unity/Packages/com.improbable.gdk.objectpooling/ObjectPooler.cs
```diff
@@ -11,13 +11,26 @@ public class ObjectPooler : MonoBehaviour
...
             objectPools = new Dictionary<GameObject, IObjectPool>();
         }
 
         public static ObjectPool<T> GetOrCreateObjectPool<T>(GameObject prefab, int prePoolCount = 0)
             where T : IPoolableObject
         {
+            if (Instance == null)
+            {
+                // Creation of the ObjectPooler component sets the Instance field
+                new GameObject("ObjectPool").AddComponent<ObjectPooler>();
```
jamiebrynes7: Given this is basically a singleton GameObject, which doesn't really have any reason to be a GameObject.. can't this just be made a classic singleton class?

peterimprobable: Generally good practice to have unused instances of the pool nested under the pool to keep the root of the hierarchy clean.
(Not sure if that's currently happening, but I'd say that'd be a nicer change than removing the gameObject entirely)

jamiebrynes7: Sure that makes sense.

I do recall hearing that keeping the objects in an nested hierarchy is worse for performance (although likely negligible here) since any change in a child object forces a re-evaluation of all the objects in its tree.

paulbalaji: It could be - but having a GameObject for it lets us place all the pooled objects as children under it

PR #77, opened 2018-12-18
Reduce bandwidth used for movement

paulbalaji:
**Contributions**: We are not currently taking public contributions - see our [contributions](https://docs.improbable.io/unity/alpha/contributing) policy. However, we are accepting issues and we do want your [feedback](https://github.com/spatialos/gdk-for-unity#give-us-feedback).

-------

#### Description
Updated bridge configurations for clients (`SimulatedPlayer` and `UnityClient`) to not checkout other clients' `ClientMovement` component - since each client only needs to care about it's own _authoritative_ `ClientMovement` component and only needs `ServerMovement` updates for checked out players.

`AdvancedEntityPipeline` now simply checks if the `ClientMovement` component is on the entity to determine whether to use an authoritative or non-authoritative prefab.

#### Tests
ran deployments with both runtime v1 and v2, and compared with old movement

|new movement before|new movement after|
|---|---|
|image|image|

_results above were with runtime v1 - bandwidth will drop even more with runtime v2 (update squashing)_

#### Documentation
n/a

#### Primary reviewers


workers/unity/spatialos.SimulatedPlayer.worker.json
```diff
@@ -14,7 +14,47 @@
     },
     "component_delivery": {
       "default": "RELIABLE_ORDERED",
-      "checkoutAllInitially": true
```
gkassabli: n00b question - what does this change? Are there any component not checked out initially in a new version?

paulbalaji: `checkoutAllInitially` means that we used to get updates for every component on checked out entities

now, we check out all components _except_ the `ClientMovement` one, because we only care about other clients' `ServerMovement` component to determine where they are

note that this is overridden for entities you have authority over - so clients will still get the `ClientMovement` component for themselves but not for other players near them

gkassabli: Makes sense. Is there any way for us to keep the default checkoutAllInitially, but only have override to false for ClientMovement that we don't have authority over?
That would make this much easier to understand - now it is difficult to parse what field is missing from overrides

paulbalaji: haha I know what you mean - I wish we could do that but according to the [docs](https://docs.improbable.io/reference/13.4/shared/worker-configuration/bridge-config#components-to-check-out-initially) we can only define overrides if `checkoutAllInitially` is false (not providing it defaults to false)

paulbalaji: a neater way to do this on runtime v2 would be to purely rely on QBI and not set the entity interest radius, which would let us avoid fiddling with bridge configs like this

but for now this solution works across both runtimes

paulbalaji:
the key changes are available on this commit:
https://github.com/spatialos/gdk-for-unity-fps-starter-project/pull/77/commits/daef7ea25e25021960e91cd6121cba1bf64f18e0

closing this PR, will bring up a new PR for current movement

PR #79, opened 2018-12-19
Split single 'look sensitivity' into seperate yaw/pitch sensitivities…

jchapmanio:
Basic update to aiming. Split single looking-about sensitivity into separate pitch/yaw sensitivities, and made pitching less sensitive on player prefab.

Also added a firing sensitivity multiplier to make aiming easier when shooting.

### DIFF
--- a/workers/unity/Assets/Fps/Resources/Prefabs/AndroidClient/Authoritative/Player.prefab
+++ b/workers/unity/Assets/Fps/Resources/Prefabs/AndroidClient/Authoritative/Player.prefab
@@ -2831,9 +2831,10 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   MovementScalar: 0.01
-  LookScalar: 0.14
+  YawMultiplier: 0.15
+  PitchMultiplier: 0.12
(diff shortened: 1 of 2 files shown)


workers/unity/Assets/Fps/Resources/Prefabs/AndroidClient/Authoritative/Player.prefab
```diff
@@ -2831,9 +2831,10 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   MovementScalar: 0.01
-  LookScalar: 0.14
+  YawMultiplier: 0.15
+  PitchMultiplier: 0.12
+  FiringLookMultiplier: 0.75
   SprintMaxAngle: 30
-  SprintDistanceThreshold: 100
```
gkassabli: Is this change intentional?

jchapmanio: I think this is a ghost serialized property. that is, one that used to exist, got serialized, then got removed in code, but the object hadn't been reserialized until now.

PR #81, opened 2018-12-20
Feature/art updates for mobile

mbirkettsmith:
Art content updates for experimental mobile branch. This includes

- Recreation of all shadergraph content in Shaderforge
- Greybox shader built around Jon's script for collapsing down tiles into large meshes to save draw calls
- Attempted to balance project around default render pipeline as opposed to HDRP
- Where possible optimising content and rebalancing to be mobile ready
- Tweaked project and quality settings to allow the same content to run correctly for both PC and Mobile
- Will just about run ok on a Galaxy S7 though I recommend dropping resolution from 1080p to 720p (requires code support). As requested is primarily balanced for S9 performance. Not we've not yet stress tested this with multiple players.
- Art content housekeeping and naming conventions pass. Content is now designed to be closer in folder and naming structure as Unreal standard naming conventions. This allows for easier cross project content sharing for the future.

N.B. previous PR for shader porting under feature/art-shader-porting should be discarded and irrelevant. This PR includes all previous updates.


workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileCollapser.cs
```diff
@@ -0,0 +1,51 @@
...
+public class TileCollapser : MonoBehaviour
+{
+    private readonly Dictionary<string, CombinedMeshAndMaterialsData> collapsedInstances =
+        new Dictionary<string, CombinedMeshAndMaterialsData>();
+
+    private void Awake()
+    {
+        for (var i = 0; i < transform.childCount; i++)
+        {
+            Debug.Log(transform.childCount);
```
firtina-improbable: This log may not be necessary (the other one in this file makes sense but may also be unnecessary)

PR #85, opened 2019-01-09
Feature/mobile mapbuilder update

jchapmanio:
- Modified Edge prefab to use Quads instead of Planes (uses just 2 tris instead of 200), updated placement logic to compensate
- Map edges now have thick walls instead of paper-thin ones ('Corner' prefab added to simplify logic of placing corners in)
- Map edges are now made up multiple pieces instead of one, potentially very very wide, piece. (Could cause collision issues
- Maps now build around the MapBuilder object instead of at origin
- All resource loading now handled in one place, and build process fails early if resource loading fails.
- Some var renaming + tidy up with current cleanup rules

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
+++ b/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
@@ -132,7 +132,11 @@ public class MapBuilder : MonoBehaviour
 
         if (centreTiles.Any(t => t == null))
         {
-            Debug.LogError("Failed to load CentreTile# resources");
+            Debug.LogError("Failed to load CentreTile resource (Expecting all the following paths to exist: " +
+                $"\t{CentreTile0}" +
+                $"\t{CentreTile1}" +
+                $"\t{CentreTile2}" +
+                $"\t{CentreTile3}");
             return false;
         }
 
--- a/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
+++ b/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Fps;
 using UnityEngine;
 using Random = UnityEngine.Random;
 
@@ -12,32 +13,33 @@ public class MapBuilder : MonoBehaviour
     public float EmptyTileChance = 0.2f;
 
     private const int TileSeparation = 36;
-    private const float GroundHeight = 0f;
-    private const float HeightOffset = 0.5f;
 
     private int groundWidth;
-    private const int WallHeight = 4;
+    private const int BlockSize = 4;
 
     private GameObject[] centreTiles;
     private GameObject[] levelTiles;
     private GameObject groundTile;
     private GameObject groundEdge;
     private GameObject surroundWall;
+    private GameObject cornerPiece;
 
     private Transform tileParentTransform;
     private Transform groundParentTransform;
     private Transform surroundParentTransform;
     private Transform cubeParentTransform;
+    private Transform spawnPointSystemTransform;
 
     private const string TileParentName = "TileParent";
     private const string GroundParentName = "GroundParent";
     private const string SurroundParentName = "SurroundParent";
-    private const string CubeParentName = "CubeParent";
+    private const string SpawnPointSystemName = "SpawnPointSystem";
 
     private const string LevelTilePath = "Prefabs/Level/Tiles";
     private const string GroundTilePath = "Prefabs/Level/Ground/Ground4x4";
     private const string GroundEdgePath = "Prefabs/Level/Ground/Edge";
     private const string SurroundPath = "Prefabs/Level/Surround/Wall";
+    private const string CornerPath = "Prefabs/Level/Surround/Corner";
 
     // Central tiles are hardcoded.
     private const string CentreTile0 = "Prefabs/Level/Tiles/Centre0";
@@ -45,20 +47,82 @@ public class MapBuilder : MonoBehaviour
     private const string CentreTile2 = "Prefabs/Level/Tiles/Centre2";
     private const string CentreTile3 = "Prefabs/Level/Tiles/Centre3";
 
+    private int groundLayers => (Layers - 1) / 4 + 1;
+
 #if UNITY_EDITOR
-    public void Build()
+    public void CleanAndBuild()
     {
+        if (TryLoadResources() == false)
+        {
+            Debug.LogError("Generation aborted due to error");
+            return;
+        }
+
+        Clean();
+
+        InitializeGroupsAndComponents();
         Random.InitState(Seed.GetHashCode());
+
+        var originalPosition = transform.position;
+        var originalRotation = transform.rotation;
+        transform.position = Vector3.zero;
+        transform.rotation = Quaternion.identity;
         PlaceTiles();
         PlaceGround();
-        PlaceSurround();
+        FillSurround();
         MakeLevelObjectStatic();
-        Debug.Log($"Finished building world of size: {groundWidth} x {groundWidth}");
+
+        transform.position = originalPosition;
+        transform.rotation = originalRotation;
+
+        var worldSize = groundLayers * TileSeparation * 4 * 2 + 4;
+        Debug.Log($"Finished building world of size: {worldSize} x {worldSize}");
     }
 
-    private void PlaceTiles()
+    private void InitializeGroupsAndComponents()
     {
-        centreTiles = new GameObject[]
+        if (!gameObject.GetComponent<TileSettings>())
+        {
+            gameObject.AddComponent<TileSettings>();
+        }
+
+        if (!GetComponentInChildren<SpawnPoints>())
+        {
+            spawnPointSystemTransform = MakeGroup(SpawnPointSystemName);
+            spawnPointSystemTransform.gameObject.AddComponent<SpawnPoints>();
+        }
+
+        groundParentTransform = MakeGroup(GroundParentName);
+
+        surroundParentTransform = MakeGroup(SurroundParentName);
+
+        tileParentTransform = MakeGroup(TileParentName);
+        tileParentTransform.gameObject.AddComponent<TileCollapser>();
+    }
+
+    private bool TryLoadResources()
+    {
+        if (!TryLoadResource(GroundTilePath, out groundTile))
+        {
+            return false;
+        }
+
+        if (!TryLoadResource(GroundEdgePath, out groundEdge))
+        {
+            return false;
+        }
+
+        if (!TryLoadResource(SurroundPath, out surroundWall))
+        {
+            return false;
+        }
+
+        if (!TryLoadResource(CornerPath, out cornerPiece))
+        {
+            return false;
+        }
+
+        centreTiles = new[]
         {
             Resources.Load<GameObject>(CentreTile0),
             Resources.Load<GameObject>(CentreTile1),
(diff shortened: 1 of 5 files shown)


workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -68,21 +132,119 @@ private void PlaceTiles()
...
+            DestroyImmediate(collision.GetComponent<MeshRenderer>());
+            DestroyImmediate(collision.GetComponent<MeshFilter>());
+        }
+    }
+
+    private void MakeCorner(int angle)
+    {
+        var rotation = Quaternion.Euler(0, angle, 0);
+        Instantiate(cornerPiece,
+            rotation * new Vector3(-groundLayers * TileSeparation * 4, 0, groundLayers * TileSeparation * 4),
```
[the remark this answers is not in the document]
jchapmanio: It could be, but it's only called four times and only in editor time so calculating it each time is pretty inconsequential.

What I could do to improve readability is make a property with a descriptive name like CornerOffset? Like,

`Vector3 CornerOffset => new Vector3(-groundLayers * TileSeparation * 4, 0, groundLayers * TileSeparation * 4)`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -68,21 +132,119 @@ private void PlaceTiles()
...
+        group.localRotation = Quaternion.identity;
+        group.localScale = Vector3.one;
+        return group;
+    }
+
+    private void FillSurround()
+    {
+        for (var i = -groundLayers; i < groundLayers; i++)
+        {
+            var offset = TileSeparation * 4 * i + TileSeparation * 2;
```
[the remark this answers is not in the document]
jchapmanio: For each bordering ground plane (which are `TileSeparation * 4` units wide/deep), it makes 1 'edge', (which is comprised of floor, wall, wall depth, and invisible collision quads) originating from the middle of the ground plane (`+ TileSeparation * 2`)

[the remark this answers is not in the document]
jchapmanio: The reason for having one edge per ground plane is that previously the walls were stretched to fill the entire map length, which can cause issues if the map is very large (e.g. floating point imprecision causing gaps due to rotation being 90.000001, or collision engine letting players stand on the centre edge of a quad on a vertical wall)

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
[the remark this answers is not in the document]
jchapmanio: in hindsight it felt like I was just bloating the class variables, so have done as you suggested and modified MakeCorner() to take in the offset as an argument

jchapmanio:
Tested locally, not in cloud yet - will do that now

jchapmanio:
Cloud test was A-OK (Jamie and I running around, no apparent issues). I've updated the level prefabs.

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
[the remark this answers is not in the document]
paulbalaji: ahh gotcha 👍

jchapmanio:
Changelog updated

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -12,53 +13,116 @@ public class MapBuilder : MonoBehaviour
...
+        var worldSize = groundLayers * TileSeparation * 4 * 2 + 4;
+        Debug.Log($"Finished building world of size: {worldSize} x {worldSize}");
     }
 
-    private void PlaceTiles()
+    private void InitializeGroupsAndComponents()
+    {
+        if (!gameObject.GetComponent<TileSettings>())
+        {
+            gameObject.AddComponent<TileSettings>();
```
jamiebrynes7: Instead of this would it be better to use a `[RequireComponent(typeof(TileSettings))]` on this Monobehaviour? 

[RequireComponent docs](https://docs.unity3d.com/ScriptReference/RequireComponent.html)

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -12,53 +13,116 @@ public class MapBuilder : MonoBehaviour
...
+        }
+
+        if (!GetComponentInChildren<SpawnPoints>())
```
jamiebrynes7: Please do a specific null check here. `(GetComponentInChildren<SpawnPoints>() == null)`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -68,21 +132,126 @@ private void PlaceTiles()
...
+        if (resource != null)
+        {
+            return true;
+        }
+
+        Debug.LogError($"Failed to load resource at Resources/{resourcePath}");
+        return false;
+    }
+
+    private Transform MakeGroup(string groupName)
```
jamiebrynes7: Nit: Rename method to `MakeChildGroup`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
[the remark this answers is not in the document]
jamiebrynes7: Might be nice to add a comment explaining this in the code :)

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -12,53 +13,116 @@ public class MapBuilder : MonoBehaviour
...
 
+    private int groundLayers => (Layers - 1) / 4 + 1;
+
 #if UNITY_EDITOR
-    public void Build()
+    public void CleanAndBuild()
     {
+        if (TryLoadResources() == false)
+        {
+            Debug.LogError("Generation aborted due to error");
```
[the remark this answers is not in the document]
jchapmanio: `TryLoadResources()` will already print out the resource that failed to load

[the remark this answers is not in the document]
jchapmanio: I could change text to "Generation aborted (See previous message for details)"?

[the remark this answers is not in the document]
jamiebrynes7: Yeah that sounds good to me 👍

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -12,53 +13,116 @@ public class MapBuilder : MonoBehaviour
...
         PlaceGround();
-        PlaceSurround();
+        FillSurround();
         MakeLevelObjectStatic();
-        Debug.Log($"Finished building world of size: {groundWidth} x {groundWidth}");
+
+        transform.position = originalPosition;
+        transform.rotation = originalRotation;
+
+        var worldSize = groundLayers * TileSeparation * 4 * 2 + 4;
```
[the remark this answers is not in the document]
jchapmanio: On reflection there are a lot of magic nums in this file...I'm going to try clarifying them all.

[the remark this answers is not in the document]
jchapmanio: I just pushed a hefty change that got rid of all the magic numbers, now all defined up front and renamed for clarity. I also updated the build complete text to give more information (playable space vs total space in 'tiles' and 'units')

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: Probably would be, yes! It does mean however that anyone already using the MapBuilder component in their project will get a compile error.

>Note that RequireComponent only checks for missing dependencies during the moment the component is added to a GameObject. Existing instances of the component whose GameObject lacks the new dependencies will not have those dependencies automatically added.

jchapmanio: Actually so long as my draw distance branch goes in, the TileSettings object is no longer necessary as its function is shifted to a singleton scriptable object.

jamiebrynes7:
Yup totally normal since the whitespace changed

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 
 - Improved the method of calculating load and FPS.
 - Upgraded to GDK for Unity version `0.1.3`
+- `MapBuilder.cs` performance and visual improvements, now generates worlds around the GameObject's position rather than at origin.
```
jamiebrynes7: Quick note - this should go under a new "Changed" heading in the Unreleased section 😄

workers/unity/Assets/Fps/Scripts/Editor/MapBuilderInspector.cs
```diff
@@ -1,37 +1,60 @@
...
-        if (GUILayout.Button("Generate Map"))
-        {
-            myTarget.Clean();
-            myTarget.Build();
+            myTarget.Seed = EditorGUILayout.TextField(new GUIContent(
+                    "Seed for Map Generator",
+                    "Different seeds produce different maps."),
+                myTarget.Seed);
+
+            myTarget.EmptyTileChance = EditorGUILayout.FloatField(new GUIContent(
```
jamiebrynes7: Should this be constrained between `0 - 1`?

workers/unity/Assets/Fps/Scripts/Editor/MapBuilderInspector.cs
```diff
@@ -1,37 +1,60 @@
...
+                    "Chance of Empty Tile",
+                    "The chance that a tile in one grid square of the world will be empty."),
+                myTarget.EmptyTileChance);
+
+            if (GUILayout.Button("Generate Map"))
+            {
+                if (myTarget.Layers < WarnLayersThreshold
+                    || EditorUtility.DisplayDialog("Generate Map Confirmation",
```
jamiebrynes7: It may be easier to read to factor this out to a method:


```
private bool GetUserConfirmation()
{
	return EditorUtility.DisplayDialog(...)
}
```

So we could just have `if (myTargetLayers < WarnLayersThreshold || GetUserConfirmation())`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -4,374 +4,431 @@
...
+        // Measurements.
+        // All sizes are 1:1 ratio in X/Z, so we just define one value to represent both axis.
+        private const int UnitsPerBlock = 4; // One textured square on the ground is a 'block'.
+        private const int UnitsPerTile = 9 * UnitsPerBlock;
+        private const int TilesPerGroundLayer = 4; // Ground layers are large quads that encompass 4x4 tiles.
+        private const int boundaryCollisionHeight = 16;
+
+        // Store the half-value as many calculations are simplified by going from -halfNumGroundLayers to halfNumGroundLayers.
+        private int HalfNumGroundLayers => (Layers - 1) / TilesPerGroundLayer + 1;
+        private int UnitsPerGroundLayer => TilesPerGroundLayer * UnitsPerTile;
```
jamiebrynes7: Nit: private properties should be `camelCase` rather than `PascalCase`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -4,374 +4,431 @@
...
-        };
 
-        if (centreTiles.Any(t => t == null))
+#if UNITY_EDITOR
+        public void CleanAndBuild()
         {
-            Debug.LogError("Central tile prefabs not loaded.");
-            return;
-        }
+            if (TryLoadResources() == false)
```
jamiebrynes7: `if (!TryLoadResources())`

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -4,374 +4,431 @@
...
+            if (levelTiles.Length <= 0)
             {
-                eulerAngles = new Vector3()
-                {
-                    y = rotation
-                }
-            },
-            tileParentTransform.transform);
-    }
+                Debug.LogError($"Failed to load resource at Resources/{LevelTilePath}");
```
jamiebrynes7: This error message doesn't really correspond to the `if` statement ?

jchapmanio: I don't follow this one, Jamie...the `if` statement checks if there were resources found, and if not, prints the error message?

Do you mean it ought to read, `Failed to load any resources at Resources/{LevelTilePath}`?

jamiebrynes7: Oh I completely missed the statement above this one! Disregard this comment 😬

PR #89, opened 2019-01-23
Add Packer Config

jamiebrynes7:
#### Description
Following up from: https://github.com/spatialos/gdk-for-unity-shared-ci/pull/16 - this adds the JSON configuration and a simple CI script to invoke the Packer.

_On the road to release automation..._ 🚶 

#### Tests
Ran the ci script. It works happily.

#### Documentation
Will be documented in our release process wiki.

Purely internal infrastructure - probably doesn't need one? 

#### Primary reviewers

### DIFF
--- a/packer.config.json
+++ b/packer.config.json
@@ -1,0 +1,27 @@
+{
+    "package_name": "gdk-for-unity-fps-starter-project",
+    "version": "0.1.3",
+    "git_packages": [
+        {
+            "clone_url": "git@github.com:spatialos/gdk-for-unity-fps-starter-project.git",
+            "clone_dir": "gdk-for-unity-fps-starter-project",
+            "exclude_paths": [
+                ".git",
+                ".github",
+                "ci",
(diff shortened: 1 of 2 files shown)


packer.config.json
```diff
@@ -0,0 +1,27 @@
...
+            "clone_dir": "gdk-for-unity-fps-starter-project",
+            "exclude_paths": [
+                ".git",
+                ".github",
+                "ci",
+                "scripts"
+            ]
+        },
+        {
+            "clone_url": "git@github.com:spatialos/gdk-for-unity.git",
```
jared-improbable: For the next iteration, it definitely makes sense to allow for a reference to a specific packer file in the cloned repo, rather than repeating things like `exclude_paths` here

jamiebrynes7: Agreed - although given the longer term plan for delivery (i.e. - UPM) we won't have the sideloaded packages and will be less of a concern.

PR #93, opened 2019-01-23
Use the new Runtime by default

paulbalaji:
#### Description
Update launch configurations to use the new Runtime.

Also updated dimensions for deployments to better match world sizes (300x300 for local/small, 900x900 for large deployments).

Local + small deployments use a 4-layer world, which has edge length of 292m. Calculated by: (4*2*36)+4. The larger deployment uses a 12-layer world, which has edge length of 868m (12*2*36)+4. This means that the spatial world sizes defined in the launch configs only need to be 300x300 and 900x900.

#### Tests
- [x] run locally
- [x] run small cloud deployment
- [x] run large cloud deployment

#### Documentation
n/a

I have now

#### Primary reviewers

### DIFF
--- a/cloud_launch_large.json
+++ b/cloud_launch_large.json
@@ -1,18 +1,36 @@
 {
-  "template": "medium_small",
+  "template": "medium_small_entity_db_v2",
   "world": {
     "chunkEdgeLengthMeters": 50,
     "snapshots": {
       "snapshotWritePeriodSeconds": 0
     },
     "dimensions": {
-      "xMeters": 1600,
-      "zMeters": 1600
+      "xMeters": 900,
+      "zMeters": 900
     },
     "legacy_flags": [
       {
         "name": "max_bridge_module_load",
         "value": "500"
+      }
+    ]
+  },
+  "load_balancing": {
+    "layer_configurations": [
+      {
+        "layer": "UnityGameLogic",
+        "rectangle_grid": {
+          "cols": 2,
+          "rows": 2
+        }
+      },
+      {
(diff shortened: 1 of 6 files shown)


cloud_launch_large.json
```diff
@@ -1,13 +1,13 @@
...
   "world": {
     "chunkEdgeLengthMeters": 50,
     "snapshots": {
       "snapshotWritePeriodSeconds": 0
     },
     "dimensions": {
-      "xMeters": 1600,
-      "zMeters": 1600
+      "xMeters": 900,
+      "zMeters": 900
```
jamiebrynes7: Any reason for this change?

cloud_launch_small.json
```diff
@@ -1,15 +1,31 @@
...
   "world": {
     "chunkEdgeLengthMeters": 50,
     "snapshots": {
       "snapshotWritePeriodSeconds": 0
     },
     "dimensions": {
-      "xMeters": 500,
-      "zMeters": 500
+      "xMeters": 300,
+      "zMeters": 300
```
jamiebrynes7: And this one too?

cloud_launch_large.json
paulbalaji: Large world has 12 layers ==> (12*2*36 + 4) ==> 868m edge length 

so the spatial world only needs to be 900x900

cloud_launch_small.json
paulbalaji: Small world has 4 layers ==> (4*2*36 + 4) ==> 292m edge length

so the spatial world only needs to be 300x300

PR #97, opened 2019-01-24
Release 0.1.4 prep

jessicafalk:
#### Description
- upgraded to 2018.3
- made sure to make FPS compatible with the master branch of the GDK
- fixed navmesh import of the simulated player coordinator

#### Tests

- [x] connect Android emulator to a local deployment
- [x] connect Android device to a local deployment
- [x] connect Mac and Windows to a cloud deployment & run fake players

#### Documentation
in progress / will be done in a separate PR

**Did you remember a changelog entry?** added it


spatialos.json
```diff
@@ -1,8 +1,8 @@
 {
     "name": "unity_gdk",
     "project_version": "0.0.1",
-    "sdk_version": "14.0-b6841-3d544-WORKER-SNAPSHOT",
+    "sdk_version": "13.5.1-b8033-b6bcd-WORKER-SNAPSHOT",
```
gkassabli: Do we want to use `13.5.1-gdk-for-unity` here?

gdk.pinned
```diff
@@ -1 +1 @@
-afa93bc4b0a96b49bbb844be5ed7d1c88d4d87a7
+master
```
gkassabli: Should we use 0.1.4 or are we going to change that later?

spatialos.json
jamiebrynes7: Yes - please make this change

gdk.pinned
jamiebrynes7: The tag isn't published yet - so has to be changed later

jessicafalk: decided to use the hash instead

PR #107, opened 2019-02-12
Deployment Manager

jessicafalk:
#### Description
Definitely read this in commit order

A service worker that uses the Platform SDK to spin up more deployments.
schema for the entity that is used to communicate between engine worker and service worker:

```
package improbable.session;

enum Status { 
    LOBBY = 1;
    RUNNING = 2;
    STOPPING = 3;
    STOPPED = 4;
} 

 // engine worker is authoritative over this one:
component Session {
    id = 1000;
    Status status = 1;
}
// deployment manager is authoritative over this one (just to ensure it has the entity checked out):
component Deployment {
    id = 1001;
}
```


workers/deployment_manager/.gitattributes
```diff
@@ -0,0 +1,63 @@
+###############################################################################
+# Set default behavior to automatically normalize line endings.
```
jared-improbable: Everything's commented out? Also, does this file work outside of the root?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            var response = deploymentServiceClient.GetDeployment(request);
+            return response.Deployment;
+        }
+
+        public static void AddDeploymentTag(string deploymentId, string projectName, string tag)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
+
+            if (!deployment.Tag.Contains(tag))
```
jared-improbable: If the tag is already in the collection, can you skip doing the update at all?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                Deployment = deployment
+            };
+
+            deploymentServiceClient.UpdateDeployment(request);
+        }
+
+        public static void RemoveDeploymentTag(string deploymentId, string projectName, string tag)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
```
jared-improbable: Same as above, but inverted :)

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+            };
+
+            deploymentServiceClient.UpdateDeployment(request);
+        }
+
+        public static void UpdateDeploymentTag(string deploymentId, string projectName, string tagPrefix, string value)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
+            var tags = deployment.Tag.ToList();
+            foreach (var tag in tags)
```
jared-improbable: Wouldn't it be better to collapse the updates down into a single call to `UpdateDeployment`?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+            var response = deploymentServiceClient.ListDeployments(request);
+            return response.ToList();
+        }
+
+        public static int GetMaxWorkerCapacity(string deploymentId, string projectName, string workerType)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
+            foreach (var workerCapacity in deployment.WorkerConnectionCapacities)
+            {
+                if (workerCapacity.WorkerType.Equals(workerType))
```
jared-improbable: Why not `==`?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+            Log.Print(LogLevel.Error, $"Couldn't find worker type {workerType} in deployment {deployment.Name}");
+            return -1;
+        }
+
+        public static int GetRemainingWorkerCapacity(string deploymentId, string projectName, string workerType)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
+            foreach (var workerCapacity in deployment.WorkerConnectionCapacities)
+            {
+                if (workerCapacity.WorkerType.Equals(workerType))
```
jared-improbable: Why not `==`?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        private string GetFileContent(string fileName)
+        {
+            string filePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName);
+            if (File.Exists(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName)))
+            {
+                Log.Print(LogLevel.Warn, $"Was not able to find file in path {filePath}");
+            }
+
+            return File.ReadAllText(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName));
```
jared-improbable: Probably a good idea to specify an explicit encoding, for example `Encoding.UTF8`

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+            string assemblyName, string deploymentNamePrefix, int numberOfDeployments, int maxPlayerCount)
+        {
+            this.projectName = projectName;
+            this.assemblyName = assemblyName;
+            this.deploymentNamePrefix = deploymentNamePrefix;
+            this.numberOfDeployments = numberOfDeployments;
+            this.clientType = clientType;
+            this.maxPlayerCount = maxPlayerCount;
+
+            var serviceAccountToken = GetFileContent("ServiceAccountToken.txt");
```
jared-improbable: It may be best to include this file as an embedded resource in the executable, rather than reading it from disk.

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+                Log.Print(LogLevel.Warn, $"Was not able to find file in path {filePath}");
+            }
+
+            return File.ReadAllText(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName));
+        }
+
+        public void ObserveDeployments()
+        {
+            // spin up the other deployments
+            var snapshotPath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, "default.snapshot");
```
jared-improbable: Same as the service account token, it may be nicer to include the snapshot as part of the binary's resources.

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+            var launchConfig = new LaunchConfig
+            {
+                ConfigJson = GetFileContent("launch_config.json"),
+            };
+
+            Log.Print(LogLevel.Info, $"Number of deployments {numberOfDeployments}");
+            for (var i = 0; i < numberOfDeployments; i++)
+            {
+                var deploymentName = $"{deploymentNamePrefix}_{i}";
+                Task.Run(() => StartDeployment(deploymentName, snapshotPath, launchConfig));
```
jared-improbable: This is an odd use of the `Task.Run` API, which isn't meant for long-running tasks like this. You may be better off using the explicit `Thread` API to start a heavyweight thread up.

workers/deployment_manager/DeploymentManager/Worker/SpatialOSReceiveHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private void OnLogMessage(LogMessageOp message)
+        {
+            Log.Print(message.Level, message.Message, connection);
+        }
+
+
+        private void OnComponentAdded(AddComponentOp<Improbable.Session.Session> componentAddOp)
+        {
+            Log.Print(LogLevel.Info, "receiving component update", connection);
```
jared-improbable: The log message is wrong, but you probably want to remove this and all the other debug prints, anyway.

workers/deployment_manager/DeploymentManager/Worker/SpatialOSReceiveHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+            Improbable.Session.Session.Data data = (Improbable.Session.Session.Data) componentAddOp.Data; // new Improbable.Session.Session.Data(Improbable.Session.SessionData.Create());
```
jared-improbable: Commented out code

workers/deployment_manager/SnapshotGenerator/Program.cs
```diff
@@ -0,0 +1,46 @@
+using System;
```
jared-improbable: Do you need the SnapshotGenerator?

workers/deployment_manager/spatialos.DeploymentManager.worker.json
```diff
@@ -0,0 +1,86 @@
...
+      "artifact_name": "DeploymentManager@Linux.zip",
+      "command": "DeploymentManager",
+      "arguments": [
+        "${IMPROBABLE_RECEPTIONIST_HOST}",
+        "${IMPROBABLE_RECEPTIONIST_PORT}",
+        "${IMPROBABLE_WORKER_ID}",
+        "${IMPROBABLE_LOG_FILE}",
+        "${IMPROBABLE_PROJECT_NAME}",
+        "${IMPROBABLE_ASSEMBLY_NAME}",
+        "UnityClient",
```
jared-improbable: Worker name wrong? (and below)

workers/deployment_manager/spatialos.DeploymentManager.worker.json
```diff
@@ -0,0 +1,86 @@
+{
+  "build": {
+    "tasks": [
+      {
+        "name": "Codegen",
+        "steps": [
+          {
+            "name": "** Load Unity to generate code **",
```
jamiebrynes7: Is this true?

workers/deployment_manager/spatialos.DeploymentManager.worker.json
```diff
@@ -0,0 +1,86 @@
...
+            "command": "echo",
+            "arguments": [ "No-op." ]
+          }
+        ]
+      },
+      {
+        "name": "build",
+        "steps": [
+          {
+            "name": "** Run ci/build-gamelogic.sh instead **",
```
jamiebrynes7: Same here

workers/deployment_manager/spatialos.DeploymentManager.worker.json
```diff
@@ -0,0 +1,86 @@
...
+        "${IMPROBABLE_RECEPTIONIST_HOST}",
+        "${IMPROBABLE_RECEPTIONIST_PORT}",
+        "${IMPROBABLE_WORKER_ID}",
+        "${IMPROBABLE_LOG_FILE}",
+        "${IMPROBABLE_PROJECT_NAME}",
+        "${IMPROBABLE_ASSEMBLY_NAME}",
+        "UnityClient",
+        "session",
+        "1",
+        "4"
```
jamiebrynes7: Would prefer that these be named arguments

workers/deployment_manager/CodeGen/CaseConversion.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+
+namespace CodeGen
+{
+    internal static class CaseConversion
+    {
+        public static string CapitalizeFirstLetter(string text)
+        {
+            return char.ToUpperInvariant(text[0]) + text.Substring(1, text.Length - 1);
+            ;
```
jamiebrynes7: Erroneous `;` ?

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
+using Google.Protobuf.Collections;
+using Google.Protobuf.WellKnownTypes;
+using Improbable.SpatialOS.ServiceAccount.V1Alpha1;
+using System;
+using System.IO;
+
+namespace ServiceAccountGenerator
+{
+
```
jamiebrynes7: Nit: New line

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+    class Program
+    {
+        static void Main(string[] args)
```
jamiebrynes7: Nit: access modifiers here

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            if (args.Length < 1)
+            {
+                Console.Out.WriteLine("Usage: <service account token path>");
```
jamiebrynes7: Is this a relative or absolute path?

jamiebrynes7: Also there appears to be more args? 

```
            var path = args[0].Trim('"');
            var projectName = args[1];
```

Would recommend using a command line lib (Mono.Options or CommandLine)

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                Environment.Exit(1);
+            }
+
+            var path = args[0].Trim('"');
+            var projectName = args[1];
+
+            Console.WriteLine(Path.GetFullPath(Path.Combine(path, "ServiceAccountToken.txt")));
+
+            var perm = new Permission
```
jamiebrynes7: Yikes - these need some sort of builder API or similar

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                Parts = { new RepeatedField<string> { "prj", projectName, "*" } },
```
jamiebrynes7: What does the wildcard do here?

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                    }
+                }
+            };
+
+            var resp = ServiceAccountServiceClient.Create().CreateServiceAccount(new CreateServiceAccountRequest
+            {
+                Name = "GDKService",
+                ProjectName = projectName,
+                Permissions = { new RepeatedField<Permission> { perm, perm2 } },
+                Lifetime = Duration.FromTimeSpan(new TimeSpan(1, 0, 0, 0)) // Let this service account live for one day
```
jamiebrynes7: How benevolent 😉

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                        Permission.Types.Verb.Read
+                    }
+                }
+            };
+
+            var resp = ServiceAccountServiceClient.Create().CreateServiceAccount(new CreateServiceAccountRequest
+            {
+                Name = "GDKService",
+                ProjectName = projectName,
+                Permissions = { new RepeatedField<Permission> { perm, perm2 } },
```
jamiebrynes7: Are the outer ` { } ` required here?

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+                Lifetime = Duration.FromTimeSpan(new TimeSpan(1, 0, 0, 0)) // Let this service account live for one day
+            });
+
+            var writer = new StreamWriter(Path.GetFullPath(Path.Combine(path, "ServiceAccountToken.txt")));
+            writer.WriteLine(resp.Token);
```
jamiebrynes7: What happens if the request fails?

workers/deployment_manager/ServiceAccountGenerator/Program.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            var resp = ServiceAccountServiceClient.Create().CreateServiceAccount(new CreateServiceAccountRequest
+            {
+                Name = "GDKService",
+                ProjectName = projectName,
+                Permissions = { new RepeatedField<Permission> { perm, perm2 } },
+                Lifetime = Duration.FromTimeSpan(new TimeSpan(1, 0, 0, 0)) // Let this service account live for one day
+            });
+
+            var writer = new StreamWriter(Path.GetFullPath(Path.Combine(path, "ServiceAccountToken.txt")));
```
jamiebrynes7: Can use a `using` block instead?

workers/deployment_manager/SnapshotGenerator/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+using System.IO;
+using Improbable;
+using Improbable.Worker;
+
+namespace SnapshotGenerator
+{
+    internal class Program
```
jamiebrynes7: Nit: Can be static?

workers/deployment_manager/SnapshotGenerator/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        private static long nextEntityId = 1;
+        private static readonly Improbable.Collections.List<Entity> Entities = new Improbable.Collections.List<Entity>();
```
jamiebrynes7: `Improbable.Collections.List` _shudders_

workers/deployment_manager/SnapshotGenerator/Program.cs
```diff
@@ -0,0 +1,46 @@
...
+                Environment.Exit(1);
+            }
+
+            var fileName = args[0].Trim('"');
+            
+            Directory.CreateDirectory(Path.GetDirectoryName(fileName));
+
+            using (var stream = new SnapshotOutputStream(fileName))
+            {
+                foreach (var entity in Entities)
```
jamiebrynes7: So.. this list is always empty..?

workers/deployment_manager/DeploymentManager/AuthenticationService/WorkerAuthenticator.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static string RetrieveDevelopmentAuthenticationToken(string projectName)
+        {
+            var request = new CreateDevelopmentAuthenticationTokenRequest
+            {
+                Description = "DAT for deployment manager",
+                Lifetime = Duration.FromTimeSpan(new TimeSpan(1, 0, 0, 0)),
+                ProjectName = projectName,
+            };
+
+            return PlayerAuthServiceClient.CreateDevelopmentAuthenticationToken(request).DevelopmentAuthenticationToken.Id;
```
jamiebrynes7: Can `CreateDevelopmentAuthenticationToken` fail?

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
```diff
@@ -0,0 +1,119 @@
+using Improbable.Worker;
+using Improbable.Worker.Alpha;
+using System.Collections.Generic;
+
+namespace DeploymentManager
+{
+    public class Connector
+    {
+        public static string LocatorHost = "locator.improbable.io";
```
jamiebrynes7: `const` ?

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
```diff
@@ -0,0 +1,119 @@
...
+        public static ushort LocatorPort = 444;
+
+        public Connection Connection { get; private set; }
+
+        private readonly string workerType;
+        private readonly Connection metaConnection;
```
jamiebrynes7: Given there are two connections - some doc strings would be 💯

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            this.metaConnection = metaConnection;
+            this.workerType = workerType;
+            this.deploymentName = deploymentName;
+        }
+
+        public string ReceivePlayerIdentityToken(string developmentAuthToken)
+        {
+            var playerIdentityTokenResponse = DevelopmentAuthentication.CreateDevelopmentPlayerIdentityTokenAsync(
+            LocatorHost,
```
jamiebrynes7: Nit: spacing is a lil wacky

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
```diff
@@ -0,0 +1,119 @@
...
+                DevelopmentAuthenticationTokenId = developmentAuthToken,
+                DisplayName = workerType,
+                PlayerId = workerType,
+                UseInsecureConnection = false,
+            }).Get();
+
+
+            if (playerIdentityTokenResponse.Status != ConnectionStatusCode.Success)
+            {
+                Log.Print(LogLevel.Fatal, playerIdentityTokenResponse.Error, metaConnection);
```
jamiebrynes7: What is this `Log`?

jamiebrynes7: Also does a fatal log automatically then kill the program?

jamiebrynes7: And we can recover from some of the status codes I think?

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
```diff
@@ -0,0 +1,119 @@
...
+        private readonly string deploymentName;
+
+        public Connector(string workerType, string deploymentName, Connection metaConnection = null)
+        {
+            this.metaConnection = metaConnection;
+            this.workerType = workerType;
+            this.deploymentName = deploymentName;
+        }
+
+        public string ReceivePlayerIdentityToken(string developmentAuthToken)
```
jamiebrynes7: `Receive` -> `Get` 

Similarly for the other functions in this class

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+using Improbable.SpatialOS.Platform.Common;
+using Improbable.SpatialOS.Snapshot.V1Alpha1;
+using System.Security.Cryptography;
+using System;
+using System.IO;
+using System.Net;
+
+namespace DeploymentManager
+{
+    public class DeploymentModifier
```
jamiebrynes7: This can be made static

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        public static string UploadSnapshot(string snapshotPath, string projectName, string deploymentName)
+        {
+            // Read snapshot.
+            var bytes = File.ReadAllBytes(snapshotPath);
+
+            if (bytes.Length == 0)
+            {
+                Console.Error.WriteLine($"Unable to load {snapshotPath}. Does the file exist?");
```
jamiebrynes7: Pretty sure if the file doesn't exist the method will throw 😉

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        public static Deployment GetDeployment(string deploymentId, string projectName)
+        {
+            var request = new GetDeploymentRequest
+            {
+                Id = deploymentId,
+                ProjectName = projectName,
+            };
+
+            var response = deploymentServiceClient.GetDeployment(request);
```
jamiebrynes7: Can this fail?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+            {
+                deployment.Tag.Add(tag);
+            }
+
+            var request = new UpdateDeploymentRequest
+            {
+                Deployment = deployment
+            };
+
+            deploymentServiceClient.UpdateDeployment(request);
```
jamiebrynes7: Can this fail?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+        public static void UpdateDeploymentTag(string deploymentId, string projectName, string tagPrefix, string value)
+        {
+            var deployment = GetDeployment(deploymentId, projectName);
+            var tags = deployment.Tag.ToList();
```
jamiebrynes7: Do we need the `ToList()`? Is the previous data struct not enumerable?

workers/deployment_manager/DeploymentManager/Log.cs
```diff
@@ -0,0 +1,78 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading;
+using Improbable.Worker;
+
+namespace DeploymentManager
+{
+    public class Log
```
jamiebrynes7: Can be static

workers/deployment_manager/DeploymentManager/Log.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        private static readonly Mutex logMutex = new Mutex();
+
+        private static StreamWriter logStream;
+
+        public static void Print(LogLevel level, string message, Connection connection = null)
+        {
+            logMutex.WaitOne();
```
jamiebrynes7: Do we need a full mutex or can we just `lock` on the StreamWriter object or similar?

workers/deployment_manager/DeploymentManager/Log.cs
```diff
@@ -0,0 +1,78 @@
...
+            Console.ForegroundColor = oldColor;
+
+            Debug.WriteLine(msg);
+
+            logStream?.WriteLine(Encoding.UTF8.GetChars(Encoding.UTF8.GetBytes(msg)));
+            logMutex.ReleaseMutex();
+
+            if (connection != null)
+            {
+                connection.SendLogMessage(level, DeploymentManager.WorkerType, message);
```
jamiebrynes7: Would it be worth limiting what we send based on the log level?

workers/deployment_manager/DeploymentManager/Log.cs
```diff
@@ -0,0 +1,78 @@
...
+                    Console.Out.WriteLine(msg);
+                    break;
+                case LogLevel.Error:
+                case LogLevel.Fatal:
+                    Console.ForegroundColor = ConsoleColor.DarkRed;
+                    Console.Error.WriteLine(msg);
+                    Environment.ExitCode = 1;
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
```
jamiebrynes7: Add a quick message here :)

workers/deployment_manager/DeploymentManager/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+            var host = args[0];
+            var port = ushort.Parse(args[1]);
+            var workerId = args.Length >= 3 ? args[2] : $"{DeploymentManager.WorkerType}-{Guid.NewGuid().ToString()}";
+            var logFileName = args.Length >= 4 ? args[3] : string.Empty;
+            var projectName = args[4];
+            var assemblyName = args[5];
+            var clientType = args[6];
+            var deploymentNamePrefix = args[7];
+            var numberOfDeployments = int.Parse(args[8]);
+            var maxPlayerCount = int.Parse(args[9]);
```
jamiebrynes7: Please use a command line parsing library 🙏

workers/deployment_manager/DeploymentManager/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+                Log.Print(LogLevel.Fatal, eventArgs.ExceptionObject.ToString());
+
+                if (eventArgs.IsTerminating)
+                {
+                    Log.Shutdown();
+                }
+            };
+
+            // Load GeneratedCode assembly
+            var generatedCode = Assembly.Load("GeneratedCode");
```
jamiebrynes7: Does it need to dynamically load it like this? Why can't it reference the DLL from the csproj?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+            WorkerAuthenticator.Init(serviceAccountToken);
+
+            // Connect to meta deployment
+            var connectionParameters = new ConnectionParameters
+            {
+                WorkerType = WorkerType,
+            };
+
+            var connector = new Connector(WorkerType, string.Empty);
+            metaConnection = connector.TryToConnect(connectionParameters, receptionistHost, receptionistPort, workerId);
```
jamiebrynes7: Alright so this is a little confusing.. in the `Connector` class what is referred to here as `metaConnection` is just `Connection` and there's a different `metaConnection` there..?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+
+            devAuthToken = WorkerAuthenticator.RetrieveDevelopmentAuthenticationToken(projectName);
+        }
+
+        private string GetFileContent(string fileName)
+        {
+            string filePath = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName);
+            if (File.Exists(Path.Combine(AppDomain.CurrentDomain.BaseDirectory, fileName)))
```
jamiebrynes7: Use the variable defined above. Also shouldn't this be `!File.Exists()`?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                Log.Print(LogLevel.Warn, $"Was not able to find file in path {filePath}");
```
jamiebrynes7: Shouldn't this early exit?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,193 @@
...
+            Log.Print(LogLevel.Info, $"Number of deployments {numberOfDeployments}");
+            for (var i = 0; i < numberOfDeployments; i++)
+            {
+                var deploymentName = $"{deploymentNamePrefix}_{i}";
+                Task.Run(() => StartDeployment(deploymentName, snapshotPath, launchConfig));
+            }
+
+            while (metaConnection.GetConnectionStatusCode() == ConnectionStatusCode.Success)
+            {
+                // to make sure the application stays alive
```
jamiebrynes7: You can just do `while(metaConnection.GetStatusCode() == ConnectionStatusCode.Success);`.. no need for the brackets :)

workers/deployment_manager/DeploymentManager/Worker/SpatialOSReceiveHandler.cs
```diff
@@ -0,0 +1,73 @@
...
+                var deployment = DeploymentModifier.GetDeployment(deploymentId, projectName);
+                DeploymentModifier.StopDeployment(deployment);
+            }
+        }
+
+        private void OnComponentUpdate(ComponentUpdateOp<Improbable.Session.Session> componentUpdate)
+        {
+            Log.Print(LogLevel.Info, "receiving component update", connection);
+            Improbable.Session.Session.Data data = new Improbable.Session.Session.Data(Improbable.Session.SessionData.Create());
+            componentUpdate.Update.ApplyTo(data);
```
jamiebrynes7: Why bother applying the update rather than just unwrapped the update directly?

workers/deployment_manager/DeploymentManager/default_launch.json
```diff
@@ -0,0 +1,60 @@
+{
+  "template": "medium_entity_db_v2",
```
jamiebrynes7: Just `medium` 😉

workers/deployment_manager/.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
```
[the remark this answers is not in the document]
jessicafalk: can probably just remove it. a remnant from taking jared's project

jared-improbable:
Just a note that a lot of comments actually are for the base `dotnet_core_worker` project, rather than specific to the Deployment Manager worker.

workers/deployment_manager/.gitattributes
jessicafalk: yeah, remnant of your project. will just delete it ;)

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
jessicafalk: still a habit from using java where `==` for strings is wrong. :P really need to stop doing that

workers/deployment_manager/spatialos.DeploymentManager.worker.json
jessicafalk: no, this is the worker name of the players that we want to observe. need to be able to pass it in so that it's not just tailored towards unity

workers/deployment_manager/SnapshotGenerator/Program.cs
jessicafalk: just to generate the initial empty snapshot for the deployment manager deployment

deployment_manager_launch.json
```diff
@@ -0,0 +1,49 @@
...
+      "snapshotWritePeriodSeconds": 0
+    },
+    "dimensions": {
+      "xMeters": 5000,
+      "zMeters": 5000
+    },
+        "legacy_flags": [
+      {
+        "name": "bridge_qos_max_timeout",
+        "value": "0"
```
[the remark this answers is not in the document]
jessicafalk: added that flag to ensure it didn't get disconnect inbetween though it should be fine now and can probably be removed

deployment_manager_launch.json
```diff
@@ -0,0 +1,49 @@
+{
+  "template": "small_entity_db_v2",
+  "world": {
+    "chunkEdgeLengthMeters": 50,
```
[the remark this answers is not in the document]
jessicafalk: maybe? couldn't find documentation ,but we still have it set for the other configs

workers/deployment_manager/publish-linux-workers.sh
```diff
@@ -0,0 +1,17 @@
...
+cd "$(dirname "$0")"
+
+dotnet build -c Release -p:Platform=x64 CodeGen/CodeGen.csproj
+dotnet build -c Release -p:Platform=x64 GeneratedCode/GeneratedCode.csproj
+dotnet build -c Release -p:Platform=x64 SnapshotGenerator/SnapshotGenerator.csproj
+dotnet publish DeploymentManager/DeploymentManager.csproj -r linux-x64 -c Release -p:Platform=x64 --self-contained
+
+
+pushd "DeploymentManager/bin/x64/Release/netcoreapp2.1/linux-x64/publish"
+  jar -cMf DeploymentManager@Linux.zip *
```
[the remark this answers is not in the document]
jessicafalk: because it's really easy to zip something with jar

workers/deployment_manager/ServiceAccountGenerator/Program.cs
jessicafalk: can this be done in a separate pr? 👼

workers/deployment_manager/ServiceAccountGenerator/Program.cs
jessicafalk: permission over ALL deployments in that project

workers/deployment_manager/publish-linux-workers.sh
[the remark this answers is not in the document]
jared-improbable: But then users have to have the JDK installed?

workers/deployment_manager/DeploymentManager/default_launch.json
jessicafalk: that one gets copy-pasted from the root folder. via the "start-deployment-manager", so would prefer to leave it like this until I rebased the base branch.

workers/deployment_manager/ServiceAccountGenerator/Program.cs
jessicafalk: it will throw an exception, e.g. `Status(StatusCode=Unauthenticated, Detail="Getting metadata from plugin failed with error: Exception occured in metadata credentials plugin.")`

jessicafalk:
that's ok ;) as soon as this PR is merged in, I'm gonna take a look at what we can move over to the base worker project

workers/deployment_manager/ServiceAccountGenerator/Program.cs
jamiebrynes7: Yes - this isn't something that should be in user code anyway!

workers/deployment_manager/DeploymentManager/AuthenticationService/WorkerAuthenticator.cs
jessicafalk: yeah, it can fail. if you try to create it for a project that you don't have a service account for

workers/deployment_manager/DeploymentManager/ConnectionService/Connector.cs
jessicafalk: it's the basic log class that Jared made.

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
jessicafalk: yeah ;) right now it just gets propagated up. we could also make it a bit more granular and catch it here?

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
jessicafalk: It's to create a copy of the list instead of using the list directly as we modify it during the iteration

workers/deployment_manager/DeploymentManager/DeploymentService/DeploymentModifier.cs
jessicafalk: oh that whole log is wrong 🤦‍♀️ 
that happens when you copy-paste from other parts

workers/deployment_manager/DeploymentManager/Log.cs
jessicafalk: probably a simple `lock` would work. gonna try

workers/deployment_manager/DeploymentManager/Log.cs
jessicafalk: maybe? right now we don't send too many messages anyway so don't think there is a point in doing that right now.

workers/deployment_manager/ServiceAccountGenerator/Program.cs
jessicafalk: fine, let's increase it to a week ;)

workers/deployment_manager/DeploymentManager/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                var options = DeploymentManagerOptions.ParseArguments(args);
+
+                // Prepare Logger
+                var logFileName = options.LogFile;
+                if (string.IsNullOrEmpty(logFileName))
+                {
+                    logFileName = Path.Combine(Environment.CurrentDirectory, $"{options.WorkerId}.log");
+                }
+
```
jared-improbable: Nit: extra line

workers/deployment_manager/SnapshotGenerator/Program.cs
jessicafalk: guess I can just remove it for now ;)

start-deployment-manager.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -e -u -x -o pipefail
+
+cd "$(dirname "$0")"
+
+if [ "$#" -ne 2 ]; then
+    echo "Wrong number of arguments. Usage: <Assembly postfix> <deployment manager project directory>"
```
joshuahuburn: `<deployment manager project directory>` is then called `LAUNCH_CONFIGURATION` later on. How come?
Also how come it's called `<Assembly postfix>` and not just assembly name?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,187 @@
...
+                var template = new DeploymentTemplate
+                {
+                    AssemblyId = options.AssemblyName,
+                    DeploymentName = deploymentName,
+                    LaunchConfig = launchConfig,
+                    ProjectName = options.ProjectName,
+                    SnapshotId = snapshotId,
+                };
+
+                while (metaConnection.GetConnectionStatusCode() == ConnectionStatusCode.Success)
```
samiwh: Is this allowed? The connection isn't generally thread safe.

start-deployment-manager.sh
```diff
@@ -0,0 +1,23 @@
...
+
+cp ${LAUNCH_CONFIGURATION} workers/deployment_manager/DeploymentManager/default_launch.json
+
+pushd "workers/deployment_manager"
+  sh publish-linux-workers.sh
+popd
+
+spatial build build-config
+spatial cloud upload "deployment_manager_${ASSEMBLY_POSTFIX}" --force
+spatial cloud launch "deployment_manager_${ASSEMBLY_POSTFIX}" deployment_manager_launch.json "deployment_manager_${ASSEMBLY_POSTFIX}" --snapshot=snapshots/empty.snapshot
```
joshuahuburn: How come we copied a launch config over to `workers/deployment_manager/DeploymentManager/default_launch.json` and then here we're using `deployment_manager_launch.json` ?

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
jared-improbable: It should be "fine", but the connection APIs are pretty unclear...

jamiebrynes7: My understanding is its _fine_ to access the connection from different threads, but no concurrent operations are allowed.

start-deployment-manager.sh
```diff
@@ -0,0 +1,23 @@
...
+echo ${ASSEMBLY_POSTFIX}
+
+cp ${LAUNCH_CONFIGURATION} workers/deployment_manager/DeploymentManager/default_launch.json
+
+pushd "workers/deployment_manager"
+  sh publish-linux-workers.sh
+popd
+
+spatial build build-config
+spatial cloud upload "deployment_manager_${ASSEMBLY_POSTFIX}" --force
```
jamiebrynes7: Given the character limits on the assembly name - I would recommend shortening the automatic prefix to lower the chance of rejection. 

I.e. - `deployment_manager_${ASSEMBLY_POSTIFX}` --> `depl_mng_${ASSEMBLY_POSTFIX}` or similar

workers/deployment_manager/ServiceAccountGenerator/ServiceAccountGenerator.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Improbable.SpatialOS.Platform" Version="13.5.1" />
```
joshuahuburn: `"13.5.1"` but the rest code specified `"13.5.0"`

start-deployment-manager.sh
jessicafalk: cause this is the launch configuration that starts the deployment of the deployment manager, not the deployments that get created by deployment manager

workers/deployment_manager/DeploymentManager/Log.cs
```diff
@@ -0,0 +1,85 @@
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+using System.Threading;
+using Improbable.Worker;
+
+namespace DeploymentManager
+{
+    public static class Log
```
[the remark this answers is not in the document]
jessicafalk: cause Jared made one and it allows us to log both to a file and to to SpatialOS

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        private readonly DeploymentManagerOptions options;
+        private readonly string devAuthToken;
+
+        private Connection metaConnection;
+
+        public DeploymentManager(DeploymentManagerOptions options)
+        {
+            this.options = options;
+
+            var serviceAccountToken = GetFileContent("ServiceAccountToken.txt");
```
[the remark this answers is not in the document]
jessicafalk: you need to make sure to ignore it and don't commit it. Our gitignore already makes sure of that. If someone commits this, it needs to be invalidated, not sure how to do this. need to check

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
jessicafalk: had an offline discussion: will look into how to refactor this so that only starting the deployments is in a separate thread and polling will be done on the main thread. that might be a better solution

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,187 @@
...
+                    Log.Print(LogLevel.Info, $"Creating deployment {deploymentName}.", metaConnection);
+                    var deploymentOperation = DeploymentModifier.CreateDeployment(template).PollUntilCompleted();
+
+                    if (deploymentOperation.IsCompleted)
+                    {
+                        Log.Print(LogLevel.Info, $"Successfully created deployment {deploymentName}.", metaConnection);
+
+                        var deploymentId = deploymentOperation.Result.Id;
+                        DeploymentModifier.AddDeploymentTag(deploymentId, options.ProjectName, "dev_login");
+                        DeploymentModifier.AddDeploymentTag(deploymentId, options.ProjectName, "ttl_1_hour");
```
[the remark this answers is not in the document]
jessicafalk: the second one: maybe, the first tag needs to be set or you can't log in to them.

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,187 @@
...
+                        var connection = ConnectToDeployment(metaConnection, deploymentOperation.Result);
+                        var handler = new SpatialOSReceiveHandler(connection, deploymentId, options.ProjectName);
+                        var time = DateTime.Now;
+
+                        while (connection.GetConnectionStatusCode() == ConnectionStatusCode.Success)
+                        {
+                            if (time < DateTime.Now)
+                            {
+                                UpdatePlayerCount(deploymentId, connection);
+                                time = DateTime.Now.AddSeconds(30);
```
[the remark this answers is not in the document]
jessicafalk: not an `ini` file, but maybe a `json` file ;)

workers/deployment_manager/DeploymentManager/Worker/DeploymentManager.cs
```diff
@@ -0,0 +1,187 @@
...
+                            }
+
+                            handler.ProcessOps();
+                        }
+
+                        Log.Print(LogLevel.Info, $"Lost connection to deployment {deploymentName}. Status: {connection.GetConnectionStatusCode()}. " +
```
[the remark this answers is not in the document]
jessicafalk: depends on the state of the deployment

workers/deployment_manager/DeploymentManager/Worker/SpatialOSReceiveHandler.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        public void ProcessOps()
+        {
+            using (var opList = connection.GetOpList(0))
+            {
+                dispatcher.Process(opList);
+            }
+        }
+
+        private void SetupHandlers()
```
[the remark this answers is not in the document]
jessicafalk: why? standard is to have public methods first and then private one

workers/deployment_manager/GeneratedCode/GeneratedCode.csproj
```diff
@@ -0,0 +1,63 @@
...
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Improbable.CoreSdk" Version="13.5.0" />
```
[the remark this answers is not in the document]
jessicafalk: we can't unless you tie it to the project, which, as we discussed, we shouldn't do. There is also no reason to do that.
What we can do: have multiple releases for the deployment manager so that people can use the release with the correct version in it

workers/deployment_manager/ServiceAccountGenerator/ServiceAccountGenerator.csproj
jessicafalk: doesn't really matter as Worker SDK & Platform SDK are completely separated. but can make them the same version

start-deployment-manager.sh
jessicafalk: might as well just change it completely and use `ASSEMBLY_NAME` instead of `ASSEMBLY_POSTFIX` as it seems to confuse people ;)

jessicafalk:
closing this in favor of the other PR that I opened

PR #115, opened 2019-02-25
Add premerge/release-qa buildkite pipelines

paulbalaji:
#### Description
- added premerge pipeline
- added release-qa pipeline
- use the `build-worker.sh` that's been moved to `gdk-for-unity-shared-ci`
- drive-by fix of deployment launcher to run `sim_player` deployment without a starting snapshot (as the snapshot was empty anyway)

#### Tests
ran on buildkite and TC

#### Documentation
n/a

#### Primary reviewers


ci/bootstrap.sh
```diff
@@ -12,7 +12,7 @@ if [ -d "${SHARED_CI_DIR}" ]; then
     rm -rf "${SHARED_CI_DIR}"
 fi
 
-git clone "${CLONE_URL}" "${SHARED_CI_DIR}"
+git clone "${CLONE_URL}" "${SHARED_CI_DIR}" --branch feature/unify-build-scripts
```
paulbalaji: this change will be removed before the final merge

ci/build-legacy.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
```
paulbalaji: this is identical to the old `ci/build.sh`, and TC will need to be updated to pick up this file instead

ci/launch.sh
```diff
@@ -0,0 +1,27 @@
...
+uploadAssembly "${PREFIX}" "${PROJECT_NAME}"
+
+markStartOfBlock "Launching deployments"
+
+dotnet run -p workers/unity/Packages/com.improbable.gdk.deploymentlauncher/.DeploymentLauncher/DeploymentLauncher.csproj -- \
+    create "${PROJECT_NAME}" "${ASSEMBLY_NAME}" \
+    "${ASSEMBLY_NAME}" cloud_launch_large.json snapshots/cloud.snapshot \
+    "${ASSEMBLY_NAME}_sim_players" cloud_launch_large_sim_players.json
+
+if [[ -n "${BUILDKITE-}" ]]; then
```
jamiebrynes7: If we never run this on TC its probably safe to remove this.. unless we except users to run this script?

ci/build-legacy.sh
paulbalaji: yes you are right - I renamed this file back before moving the ci/build.sh to shared-ci

ci/launch.sh
paulbalaji: Nope. Users can use this as an idea of how to use the deployment launcher through scripts if they want, but we shouldn't expect users to run any of our ci/ scripts

paulbalaji:
> If we are removing the sim players snapshot - what snapshot is the deployment started with now?

The `_sim_player` deployment is started up without a snapshot now, as it was empty before anyway. This works as normal in testing

ci/build.sh
```diff
@@ -10,11 +10,11 @@ ci/bootstrap.sh
```
jamiebrynes7: File permissions on this got rekt. Should be reverted to `100755`

ci/launch.sh
```diff
@@ -0,0 +1,28 @@
...
+
+markStartOfBlock "Launching deployments"
+
+dotnet run -p workers/unity/Packages/com.improbable.gdk.deploymentlauncher/.DeploymentLauncher/DeploymentLauncher.csproj -- \
+    create "${PROJECT_NAME}" "${ASSEMBLY_NAME}" \
+    "${ASSEMBLY_NAME}" cloud_launch_large.json snapshots/cloud.snapshot \
+    "${ASSEMBLY_NAME}_sim_players" cloud_launch_large_sim_players.json
+
+CONSOLE_URL="https://console.improbable.io/projects/${PROJECT_NAME}/deployments/${ASSEMBLY_NAME}/overview"
+CONSOLE_URL_SIM_PLAYERS="https://console.improbable.io/projects/${PROJECT_NAME}/deployments/${ASSEMBLY_NAME}_sim_players/overview"
```
jamiebrynes7: This is way better than the regex approach 😂

ci/build.sh
paulbalaji: good spot, fixed 👍

PR #118, opened 2019-02-28
Feature/aaa mapgenerator update

jchapmanio:
This update gives more control over the types and locations of tiles in the generated world.

NB. This is branched from and requires content from `feature/aaa-level-update` branch

NEW
`TileTypeCollection.cs` defines a ScriptableObject that can hold a list of prefabs, a % chance of empty tile, and a representation colour used by `TileTypeVolume`s.

`TileTypeVolume.cs` a component that uses a Box Collider to define a region in which a certain `TileTypeCollection` should be used. Can auto-rename the gameObject attached to for organisational purposes.

Volumes should be placed within a prefab and referenced by the `MapBuilderSettings`.

CHANGED
`MapBuilder.cs` / `MapBuilderSettings.cs` 
- now takes in a 'TileTypeVolumes' prefab reference which is instantiated before the map is generated, then destroyed afterward. 
- can take a 'Default collection' `TileTypeCollection` reference, and will spawn tiles from this collection if no volume is found at a location.
- no longer has a default 'empty tile chance'. This is driven entirely within the `TileTypeCollection` object.

image
Example output


workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -180,6 +166,7 @@ private bool TryLoadResource(string resourcePath, out GameObject resource)
             return false;
         }
 
+
```
paulbalaji: nit: extra line

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -57,8 +53,7 @@ public MapBuilder(MapBuilderSettings mapBuilderSettings, GameObject gameObject)
 
         public void CleanAndBuild(
             int worldLayers = 4,
-            string seed = "SpatialOS GDK for Unity",
-            float emptyTileChance = 0.2f)
+            string seed = "SpatialOS GDK for Unity")
```
jamiebrynes7: Nit: This can probably go all in one line now

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -400,6 +394,26 @@ private void MakeLevelObjectStatic()
...
+                {
+                    continue;
+                }
+
+                return volume.TypeCollection.GetRandomTile();
+            }
+
+            return mapBuilderSettings.DefaultTileType == null
+                ? null
+                : mapBuilderSettings.DefaultTileType.GetRandomTile();
```
jamiebrynes7: If the default is null, we return null, else we return a random tile? 

That doesn't make a lot of sense? Surely we either return the default tile type always or return a random tile?

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -477,7 +491,17 @@ public static string GetWorldSizeFlag(Connection connection)
                 levelInstance.transform.rotation = workerTransform.rotation;
 
                 var mapBuilder = new MapBuilder(mapBuilderSettings, levelInstance);
+
+                var volumesPrefab = mapBuilderSettings.WorldTileVolumes == null
+                    ? null
+                    : MonoBehaviour.Instantiate(mapBuilderSettings.WorldTileVolumes);
```
jamiebrynes7: Why `MonoBehaviour.Instantiate` and `MonoBehaviour.Destroy` instead of `GameObject.Instantiate` and `GameObject.Destroy` ?

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
```diff
@@ -0,0 +1,23 @@
+﻿using UnityEngine;
+
+namespace Fps
+{
+    [CreateAssetMenu(fileName = "New Prefab Collection", menuName = "Improbable/Create Prefab Collection")]
```
jamiebrynes7: Asset and menu name are less specific than what this is.

`Prefab Collection` vs `Tile Collection`

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class TileTypeCollection : ScriptableObject
+    {
+        [SerializeField] private GameObject[] Tiles;
+        [Range(0f, 1f)] public float ChanceOfEmptyTile;
+        public Color DisplayColor = Color.blue;
+
+        public GameObject GetRandomTile()
+        {
+            var rand = Random.value;
+            if (ChanceOfEmptyTile == 0f || rand > ChanceOfEmptyTile)
```
jamiebrynes7: Is the first part of this statement strictly necessary?

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
```diff
@@ -0,0 +1,52 @@
+﻿using Fps;
+using UnityEngine;
+
+[RequireComponent(typeof(BoxCollider))]
+public class TileTypeVolume : MonoBehaviour
```
jamiebrynes7: Needs namespace

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private void OnDrawGizmos()
+    {
+        if (TypeCollection == null)
+        {
+            return;
+        }
+
+        if (boxCollider == null)
```
jamiebrynes7: Can `OnDrawGizmos` happen before `Awake` ?

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public GameObject GetRandomTile()
+    {
+        if (TypeCollection == null)
+        {
+            return null;
+        }
+
+        return TypeCollection.GetRandomTile();
```
jamiebrynes7: This can just be `TypeCollection?.GetRandomTile()` 😄

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: DefaultTileType is a collection of 1 or more tiles, so it's,

If the default collection is null, we return null, otherwise we return a random tile from the default collection.

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: No reason; should they be `GameObject.` instead?

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
jchapmanio: Good catch, some legacy stuff from the script's old lifestyle as a generic prefab collection there.

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
jchapmanio: It is; if `rand == 0f`, and `ChanceOfEmptyTile == 0f`, then `rand > ChanceOfEmptyTile` is false and so it makes an empty tile despite ChanceOfEmptyTile being zero. There might be a better way to do this, though.

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
jchapmanio: Yes, in editor mode `OnDrawGizmos` is called but `Awake` isn't.

jchapmanio: (Unless `[ExecuteInEditMode]` attribute is added to the class, but that doesn't make sense here)

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
jchapmanio: Ahh good 'ol null propagations. I need to get used to using them!

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: Ohh wait, could this be a case for null propagation? `return mapBuilderSettings.DefaultTileType?.GetRandomTile();`?

jamiebrynes7: Yes - also can we rename `DefaultTileType` to `TileCollection` or similar then? 

If there are no overrides, the name Default doesn't make a ton of sense

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jamiebrynes7: Not sure - I've never seen those APIs. You may actually want `Object.Destroy()` as that's the base implementation

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
jamiebrynes7: Why not `rand >= ChanceOfEmtpyTile` then? 

Also `==` operations on floating points are always a little funky

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeVolume.cs
jamiebrynes7: Cool - just wanted to double check 😄

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: If I understand correctly, then there are overrides. If a space in the world is occupied by a TileTypeVolume, then that volume overrides the default tile type value.

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jchapmanio: typically you'd be calling Destroy or Instantiate from within a Monobehaviour script. We could change it to Object, but I don't understand why one is better than the other?

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
jchapmanio: aha, because if `rand == 1.0f` and `ChanceOfEmptyTile == 1.0f`, then it will not make an empty tile - and it should!

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jamiebrynes7: Just consistency across the codebase really

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jamiebrynes7: Ah yes you are correct, I wasn't looking at the broader context.

Can we rename to make it clear its a collection of tiles? `DefaultTileTypes`

jchapmanio: 👍 I'll name it `DefaultTileTypeCollection`

workers/unity/Assets/Fps/Scripts/GameLogic/WorldTiles/TileTypeCollection.cs
```diff
@@ -0,0 +1,23 @@
+﻿using UnityEngine;
+
+namespace Fps
+{
+    [CreateAssetMenu(fileName = "New TileTypeCollection", menuName = "Improbable/Create Tile Type Collection")]
```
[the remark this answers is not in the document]
jchapmanio: Everything else is Fps atm, including the `MapBuilder` script. True, it's not FPS-specific - but it is the only place we use it currently.

workers/unity/Assets/Fps/Config/TileTypes/DefaultTiles.asset
```diff
@@ -12,12 +12,12 @@ MonoBehaviour:
...
+  tiles:
   - {fileID: 1853307461053178640, guid: f4b0c5a26e0dde042af81ff54b6149bb, type: 3}
   - {fileID: 1853307461053178640, guid: 77073bec1cfdc74448a1433b0c2eb0a0, type: 3}
   - {fileID: 1853307461053178640, guid: d6da8e0d7a1d5654e8b23301a33af160, type: 3}
   - {fileID: 1853307461053178640, guid: 108e84c74f708804abfc1ac92731b83d, type: 3}
   - {fileID: 1853307461053178640, guid: 46057f07999102d459eb1e5a48c6347f, type: 3}
   - {fileID: 1853307461053178640, guid: 7c12f1fe6891d5747b06b5c2cc0a9650, type: 3}
-  ChanceOfEmptyTile: 0.05
-  DisplayColor: {r: 1, g: 1, b: 1, a: 0.5019608}
+  chanceOfEmptyTile: 0
```
zeroZshadow: Is it correct that this chance has changed?

jchapmanio: oh, good spot. Not, it is not correct!

PR #119, opened 2019-02-28
Core 0.2 upgrade

samiwh:
#### Description
Upgrading to match the core refactor.
Mostly rename changes.
The game object creator now has to track what objects it creates.
A couple of null checks from slight ordering changes.


workers/unity/Packages/com.improbable.gdk.guns/Behaviours/FirstPersonRecoil.cs
```diff
@@ -39,6 +39,11 @@ private static Vector3 CalculateNewVelocity(GunSettings.FirstPersonRecoilSetting
 
         private void LateUpdate()
         {
+            if (aimRecoil == null || hipRecoil == null)
```
zeroZshadow: What is this change for?

workers/unity/Packages/com.improbable.gdk.guns/Behaviours/GunIK.cs
```diff
@@ -55,6 +54,12 @@ private void Awake()
         private void OnAnimatorIK(int layer)
         {
             var recoilOffset = GetRecoilOffset();
+
+            if (gunSocket.Gun == null)
```
zeroZshadow: What is this change for?

workers/unity/Packages/com.improbable.gdk.guns/Behaviours/FirstPersonRecoil.cs
samiwh: As said in the description this is to deal with an ordering change made in the new core stuff. Previously this was assuming the monobehaviour would enable before another, which was never garunteed.

PR #122, opened 2019-03-03
Updating build config & adding checks to SimulatedPlayerWorkerCoordinatorConnector

jessicafalk:
#### Description
- Updated build configuration to work with the new build system UI
- Added additional checks to the `SimulatedPlayerWorkerCoordinator` to ensure it doesn't spam errors, if play mode is exited at the wrong time. 

#### Tests
tested locally in editor

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/SetupLogic/SimulatedPlayerCoordinatorWorkerConnector.cs
+++ b/workers/unity/Assets/Fps/Scripts/SetupLogic/SimulatedPlayerCoordinatorWorkerConnector.cs
@@ -64,9 +64,10 @@ public class SimulatedPlayerCoordinatorWorkerConnector : WorkerConnectorBase
         }
     }
 
-    private void OnDestroy()
+    public override void Dispose()
     {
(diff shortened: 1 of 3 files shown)


workers/unity/Assets/Fps/Scripts/SetupLogic/SimulatedPlayerCoordinatorWorkerConnector.cs
```diff
@@ -48,20 +47,29 @@ protected override async void HandleWorkerConnectionEstablished()
...
 
-            StartCoroutine(MonitorSimulatedPlayers());
+            if (Application.isPlaying)
+            {
+                StartCoroutine(MonitorSimulatedPlayers());
+            }
         }
     }
 
+    public override void Dispose()
```
jessicafalk: decided to move it to `Dispose` instead of `OnDestroy`to keep it consistent with the other connectoes where we overrode `Dispose` instead of `OnDestroy` - also means that this branch is not dependent on the other branch in the gdk repo anymore

PR #124, opened 2019-03-06
Fix release-qa pipeline by passing in region code to deployment launcher

paulbalaji:
#### Description
pass in region code

as we'll be in the UK for QA, we will use the EU region

#### Tests
buildkite

#### Documentation
n/a

#### Primary reviewers


jamiebrynes7:
Insert Brexit joke here

PR #126, opened 2019-03-08
Map Builder Visualisation now auto-creates volumes prefab using MapBu…

jchapmanio:
Spawns the tile-type volumes prefab before building the map through the Map Builder Visualisation tool, then removes them again. (Previously had to manually place the prefab into the scene)


workers/unity/Assets/Fps/Scripts/Editor/MapBuilderVisualisationWindow.cs
```diff
@@ -70,7 +70,23 @@ public void OnGUI()
...
 
+                    var volumesPrefab = mapBuilderSettings.WorldTileVolumes == null
+                        ? null
+                        : Instantiate(mapBuilderSettings.WorldTileVolumes);
+
                     mapBuilder.CleanAndBuild(layerCount, seed);
+
+                    if (volumesPrefab != null)
+                    {
+                        if (Application.isPlaying)
```
zeroZshadow: There is a helper utility for that that is used in the map builder as well. it will do this exact `if else` bit.

jchapmanio: Hmm I can't find it

jessicafalk: this one: https://github.com/spatialos/gdk-for-unity/blob/master/workers/unity/Packages/com.improbable.gdk.core/Utility/UnityObjectDestroyer.cs

jchapmanio:
UnityObjectDestroyer wasn't used in MapBuilder.cs, but it is now.

PR #127, opened 2019-03-11
Feature/art updates

mbirkettsmith:
Substantial character updates, some performance improvements and bugfixes

Latest reworked character model and textures. Better LOD setup and fewer triangles for first person mesh.
Overhauled entire character animation setup, now has much better animations that have been reintegrated by Pete.
Added in new ragdoll for character from Adam

Bugfix to remove flickering SSAO on PC

Updated FOV for PC and mobile. Will be of particular benefit for readability and performance on mobile. 

Some environmental shader bugfixes


zeroZshadow:
I've noticed that there is a small bug where the gun doesnt show untill you move

PR #137, opened 2019-03-14
Clean up worker connectors + make map building deferred over frames

jamiebrynes7:
#### Description
This PR has a few changes: 
- Drive-by cleanup of some old worker configurations.
- Made map building deferred over multiple frames instead of blocking. This stops the `SimulatedPlayerCoordinator` workers from falling over at startup and allow them to maintain a reasonable FPS such that they can respond to pings.
	- Note that this does have consequences on other parts of the codebase (i.e. - cannot spawn until map is built). There is still an unresolved question of what happens if a UnityGameLogic worker goes down and starts to checkout existing players _before_ the map is built, but this is an edge case (and I expect that since we are client side auth, it will _probably_ be okay). 
- Cleaning up the structure of the worker connectors (still work to do, but they are more aligned). We now have the following class structure:
```
WorkerConnectorBase
  - ClientWorkerConnector
  - GameLogicWorkerConnector
  - SimulatedPlayerCoordinatorConnector
  - MobileWorkerConnectorBase
    - AndroidWorkerConnector
	- iOSWorkerConnector 
```

Ultimately the client and mobile ones can probably be unified as there is a lot in common with them. Equally if the worker abstraction and the connectors are likely to change in the near-ish future (which it looks like they are) it may not be worth going that far.

#### Tests
- [x] Tested local deployment
- [ ] Test Android
- [ ] Test iOS
- [ ] Test cloud deployment
#### Documentation
Needs a changelog.

#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/SetupLogic/WorkerConnectorBase.cs
+++ b/workers/unity/Assets/Fps/Scripts/SetupLogic/WorkerConnectorBase.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections;
 using System.Threading.Tasks;
 using Improbable.Gdk.Core;
@@ -12,7 +13,7 @@ namespace Fps
 
         [SerializeField] protected MapBuilderSettings MapBuilderSettings;
 
-        public GameObject LevelInstance;
+        [NonSerialized] internal GameObject LevelInstance;
 
         protected abstract string GetWorkerType();
 
--- a/workers/unity/Assets/Fps/Scripts/SetupLogic/WorkerConnectorBase.cs
+++ b/workers/unity/Assets/Fps/Scripts/SetupLogic/WorkerConnectorBase.cs
@@ -1,3 +1,4 @@
+using System.Collections;
 using System.Threading.Tasks;
 using Improbable.Gdk.Core;
 using Improbable.Worker.CInterop;
@@ -11,7 +12,7 @@ namespace Fps
 
         [SerializeField] protected MapBuilderSettings MapBuilderSettings;
 
-        protected GameObject levelInstance;
+        public GameObject LevelInstance;
 
         protected abstract string GetWorkerType();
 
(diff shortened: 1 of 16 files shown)


jchapmanio:
`MapBuilderVisualisationWindow.cs` should be updated to accommodate the MapBuilder changes; currently using it (SpatialOS -> Map Builder) will not generate anything.

jchapmanio:
Errors when launching in Development scene:

```
UnassignedReferenceException: The variable LevelInstance of GameLogicWorkerConnector has not been assigned.
You probably need to assign the LevelInstance variable of the GameLogicWorkerConnector script in the inspector.

UnassignedReferenceException: The variable LevelInstance of SimulatedPlayerCoordinatorWorkerConnector has not been assigned.
You probably need to assign the LevelInstance variable of the SimulatedPlayerCoordinatorWorkerConnector script in the inspector.

UnassignedReferenceException: The variable LevelInstance of ClientWorkerConnector has not been assigned.
You probably need to assign the LevelInstance variable of the ClientWorkerConnector script in the inspector.
```

jamiebrynes7:
Huh I tested that one locally - let me double check

workers/unity/Assets/Fps/Scripts/SetupLogic/WorkerConnectorBase.cs
```diff
@@ -11,7 +12,7 @@ public abstract class WorkerConnectorBase : DefaultWorkerConnector
 
         [SerializeField] protected MapBuilderSettings MapBuilderSettings;
 
-        protected GameObject levelInstance;
+        public GameObject LevelInstance;
```
jchapmanio: I'm not clear on what the intended usage is of the public LevelInstance reference. Is it supposed to be hidden in the inspector? If not, what object is to be referenced by it?

jamiebrynes7: Its used by the map builder - so has to be public (or internal I guess)

jchapmanio: Does it need to be accessed from the inspector? If not, a `[HideInInspector]` attribute would be preferable to avoid people thinking it's something they need to worry about, or if it doesn't need to be serialized, `[NonSerialized]`

jamiebrynes7: I don't think its accessed from the inspector - will add the attribute 👍

jchapmanio:
Not tested outside of editor, but LGTM

jchapmanio:
I will not approve though as I cannot review connection related changes.

zeroZshadow:
Needs TileEnabler fixes, will add

PR #151, opened 2019-04-03
Deployment launcher refactor part 1 - Pull out configurations + commands

jamiebrynes7:
#### Description

* Created objects to represent: a deployment configuration for launching, deployment info received from the console, and assembly configuration for uploading.
* Factored out commands to take these as arguments. Note that I planned here that I will refactor the dotnet project to have named arguments.

**There are some TODOs in this change - that is okay for now**

--- 

Notes on `WrappedTask<TResult, TContext>`: This allows us to bundle the context for a task, the task itself, and the cancellation source into one object. The reason we bundle the context is that whoever called the command can print useful messages for a user when the task finishes without holding onto a copy of the source object themselves.

The fields inside the UI may have changed in the meantime - hence the `DeepCopy` on the configurations. 

Note that this isn't required for the `DeploymentInfo` as we receive this off the "wire" and is immutable (getter only properties). The reason we can't make the other configs immutable is that the UI will be writing to a copy of these when the user changes the fields.

#### Tests
None yet.

#### Documentation
Will do this at the end.

#### Primary reviewers
@jessicafalk @jared-improbable

### DIFF
--- a/workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Deployment.cs
+++ b/workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Deployment.cs
@@ -19,7 +19,6 @@ namespace Improbable.Gdk.DeploymentManager.Commands
             var source = new CancellationTokenSource();
             var token = source.Token;
 
-
             // TODO: Allow for no snapshot
             var args = new[]
             {
--- a/workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Deployment.cs
+++ b/workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Deployment.cs
@@ -1,0 +1,129 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Improbable.Gdk.Tools;
+
+namespace Improbable.Gdk.DeploymentManager.Commands
+{
+    public static class Deployment
+    {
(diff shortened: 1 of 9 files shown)


workers/unity/Packages/com.improbable.gdk.deploymentlauncher/DeploymentConfigurations.cs
```diff
@@ -0,0 +1,110 @@
...
+        ///     The project to upload this assembly to.
+        /// </summary>
+        public string ProjectName { get; private set; }
+
+        /// <summary>
+        ///     The name of this assembly.
+        /// </summary>
+        public string AssemblyName { get; private set; }
+
+        public AssemblyConfig DeepCopy()
```
paulbalaji: xml docs

Also why is DplConfig.DeepCopy internal but this public?

workers/unity/Packages/com.improbable.gdk.deploymentlauncher/Commands/Deployment.cs
```diff
@@ -0,0 +1,129 @@
...
+        private static string DeploymentLauncherProjectPath = Path.GetFullPath(Path.Combine(
+            Tools.Common.GetPackagePath("com.improbable.gdk.deploymentlauncher"),
+            ".DeploymentLauncher/DeploymentLauncher.csproj"));
+
+        public static WrappedTask<bool, DeploymentConfig> LaunchAsync(DeploymentConfig config)
+        {
+            var source = new CancellationTokenSource();
+            var token = source.Token;
+
+
```
paulbalaji: nit: extra line

PR #181, opened 2019-05-13
Adding callbacks for losing player

jessicafalk:
#### Description
Especially on mobile it can happen that SpatialOS thinks a worker is still connected, but the server-worker heartbeats fail at some point. When that happens we delete the player entity, but not disconnect the client leaving the client in a bad state. This PR creates some callbacks to disconnect whenever we lose the player entity and go back to the start screen.

#### Tests
running in editor by deleting player entities via the inspector.

#### Documentation
How is this documented (for example: release note, upgrade guide, feature page, in-code documentation)?

#### Primary reviewers

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
@@ -1,3 +1,5 @@
+using UnityEngine;
+
 namespace Fps
 {
     public class DefaultPlayState : DefaultState
@@ -9,6 +11,7 @@ namespace Fps
         public override void StartState()
         {
             Blackboard.ClientConnector.Worker.OnDisconnect += WorkerOnDisconnect;
+            Blackboard.ClientConnector.OnLostPlayerEntity += LostPlayerEntity;
             Manager.InGameManager.Timer.SetActive(false);
             Manager.ShowGameView();
         }
@@ -16,9 +19,15 @@ namespace Fps
         public override void ExitState()
         {
             Blackboard.ClientConnector.Worker.OnDisconnect -= WorkerOnDisconnect;
+            Blackboard.ClientConnector.OnLostPlayerEntity -= LostPlayerEntity;
         }
 
         private void WorkerOnDisconnect(string reason)
+        {
+            Owner.SetState(new DisconnectedState(Manager, Owner));
+        }
+
+        private void LostPlayerEntity()
         {
             Owner.SetState(new DisconnectedState(Manager, Owner));
         }
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
@@ -1,5 +1,3 @@
-using UnityEngine;
-
 namespace Fps
 {
     public class DefaultPlayState : DefaultState
--- a/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
+++ b/workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
@@ -27,6 +27,9 @@ namespace Fps
 
         private void LostPlayerEntity()
         {
+            Manager.ShowFrontEnd();
+            ScreenManager.SwitchToDefaultScreen();
+            Animator.SetTrigger("Disconnected");
             Owner.SetState(new DisconnectedState(Manager, Owner));
         }
     }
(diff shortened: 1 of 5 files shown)


workers/unity/Assets/Fps/Scripts/SetupLogic/OneTimeInitialisation.cs
```diff
@@ -23,6 +23,7 @@ private static void Init()
 
             // Setup template to use for player on connecting client
             PlayerLifecycleConfig.CreatePlayerEntityTemplate = FpsEntityTemplates.Player;
+            PlayerLifecycleConfig.MaxNumFailedPlayerHeartbeats = 5;
```
jessicafalk: increased that number to reduce the change of this heartbeating failing first.
but open to revert that change if people think that's not necessary

workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DefaultPlayState.cs
```diff
@@ -1,3 +1,5 @@
+using UnityEngine;
```
jessicafalk: note to self: need to remove the import as soon as unity finished building

workers/unity/Assets/Fps/Scripts/GameLogic/StateMachine/Default/DisconnectedState.cs
```diff
@@ -8,6 +10,7 @@ public DisconnectedState(UIManager manager, ConnectionStateMachine owner) : base
 
         public override void StartState()
         {
+            UnityObjectDestroyer.Destroy(Blackboard.ClientConnector.gameObject);
```
zeroZshadow: This can be a normal `UnityEngine.Object.Destroy` if the gameObject exists for 1 frame or more, as the code is not run in editor mode.

workers/unity/Assets/Fps/Scripts/SetupLogic/AdvancedEntityPipeline.cs
```diff
@@ -6,94 +6,104 @@
...
-public class AdvancedEntityPipeline : IEntityGameObjectCreator
+namespace Fps
 {
-    private const string GameobjectNameFormat = "{0}(SpatialOS {1}, Worker: {2})";
-    private const string WorkerAttributeFormat = "workerId:{0}";
-    private const string PlayerMetadata = "Player";
+    public class AdvancedEntityPipeline : IEntityGameObjectCreator
+    {
+        private const string GameobjectNameFormat = "{0}(SpatialOS {1}, Worker: {2})";
+        private const string WorkerAttributeFormat = "workerId:{0}";
```
zeroZshadow: Don't we have a build in GDK helper for this now?

workers/unity/Assets/Fps/Scripts/SetupLogic/AdvancedEntityPipeline.cs
```diff
@@ -6,94 +6,104 @@
...
+        public void OnEntityRemoved(EntityId entityId)
+        {
+            if (!gameObjectsCreated.TryGetValue(entityId, out var go))
+            {
+                fallback.OnEntityRemoved(entityId);
+                return;
+            }
+
+            // Trigger a callback when authoritative player gets removed
+            if (go.GetComponent<FpsDriver>() != null)
```
zeroZshadow: This looks a bit hacky, is there no better way to check if this entity was the player authoritative one?

workers/unity/Assets/Fps/Scripts/SetupLogic/ClientWorkerConnector.cs
```diff
@@ -109,6 +112,12 @@ protected override void HandleWorkerConnectionEstablished()
             base.HandleWorkerConnectionEstablished();
         }
 
+        private void RemovingAuthoritativePlayer()
+        {
+            Debug.LogError($"Removing authoritative player.");
```
zeroZshadow: This message doesn't feel like an error. Might need some clarification in the message as to why this is bad.
"Player entity was removed while connected, disconnecting" ?

workers/unity/Assets/Fps/Scripts/SetupLogic/AdvancedEntityPipeline.cs
jessicafalk: after checking , the entity it was linked to is already destroyed. So we can't use this without refactoring some other stuff. Will leave this for now as it is

workers/unity/Assets/Fps/Scripts/SetupLogic/AdvancedEntityPipeline.cs
```diff
@@ -39,7 +39,7 @@ public class AdvancedEntityPipeline : IEntityGameObjectCreator
         {
             this.worker = worker;
             this.fallback = fallback;
-            workerIdAttribute = string.Format(WorkerAttributeFormat, worker.WorkerId);
+            workerIdAttribute = EntityTemplate.GetWorkerAccessAttribute(worker.WorkerId);
```
zeroZshadow: This means `WorkerAttributeFormat` can now be removed.

PR #184, opened 2019-05-21
Pin shared-ci version

paulbalaji:
#### Description

This is defensive of a future change in the GDK to stop copying schema files around. Said change means that we no longer have to run `spatial prepare-for-run` in the worker build step. 

As different projects/branches now require different commits from the shared-ci repo, we need to pin which version we use.

#### Tests

- [x] copied across the pinning/cloning code from https://github.com/spatialos/gdk-for-unity/pull/953/commits/614055ea1a7aa0e35588005a125c29daefb6f4a2

#### Documentation

n/a

#### Primary reviewers

### DIFF
--- a/ci/bootstrap.sh
+++ b/ci/bootstrap.sh
@@ -5,14 +5,21 @@ cd "$(dirname "$0")/../"
 
 SHARED_CI_DIR="$(pwd)/.shared-ci"
 CLONE_URL="git@github.com:spatialos/gdk-for-unity-shared-ci.git"
+PINNED_SHARED_CI_VERSION=$(cat ./ci/shared-ci.pinned)
 
 # Clone the HEAD of the shared CI repo into ".shared-ci"
 
-if [[ -d "${SHARED_CI_DIR}" ]]; then
-    rm -rf "${SHARED_CI_DIR}"
-fi
+rm -rf "${SHARED_CI_DIR}"
 
-git clone --verbose --depth 1 "${CLONE_URL}" "${SHARED_CI_DIR}"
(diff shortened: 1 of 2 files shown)


ci/bootstrap.sh
```diff
@@ -5,14 +5,21 @@ cd "$(dirname "$0")/../"
...
 SHARED_CI_DIR="$(pwd)/.shared-ci"
 CLONE_URL="git@github.com:spatialos/gdk-for-unity-shared-ci.git"
+PINNED_SHARED_CI_VERSION=$(cat ./ci/shared-ci.pinned)
 
 # Clone the HEAD of the shared CI repo into ".shared-ci"
 
-if [[ -d "${SHARED_CI_DIR}" ]]; then
-    rm -rf "${SHARED_CI_DIR}"
-fi
+rm -rf "${SHARED_CI_DIR}"
```
zeroZshadow: Why does this no longer check if the directory exists before removing?

jessicafalk: I guess it doesn't really matter? either it exists and we remove it or it doesn't exist and we simply try to remove it

paulbalaji: ^this was my thinking too

zeroZshadow: I assume the test was there for a reason previously, so I'm not sure why it was removed.

paulbalaji: there was no reason other than the outcome being same with or without the test

added it back as it is a harmless check anyway (updated blank project PR too)

PR #198, opened 2019-06-11
Fix mapbuilder in editor

zeroZshadow:
#### Description
Refactored some bits of the map builder to work in both the editor and play mode again.

#### Tests
Testing in a stand alone build atm

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
+++ b/workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
@@ -3,7 +3,6 @@ using System.Collections;
 using System.Collections.Generic;
 using Improbable.Gdk.Core;
 using UnityEngine;
-using UnityEngine.Profiling;
 using UnityEngine.SceneManagement;
 using Object = UnityEngine.Object;
 using Random = System.Random;
@@ -18,7 +17,7 @@ namespace Fps
         private int layers;
         private Random random;
         private Scene tempScene;
-        private Scene activeScene;
+        private Scene mainScene;
 
         private GameObject groundTile;
         private GameObject groundEdge;
@@ -67,8 +66,12 @@ namespace Fps
 
             layers = worldLayers;
             random = new Random(seed.GetHashCode());
-            tempScene = SceneManager.CreateScene(parentGameObject.name, new CreateSceneParameters(LocalPhysicsMode.None));
-            activeScene = SceneManager.GetActiveScene();
+            if (Application.isPlaying)
+            {
+                tempScene = SceneManager.CreateScene(parentGameObject.name,
+                    new CreateSceneParameters(LocalPhysicsMode.None));
+                mainScene = SceneManager.GetActiveScene();
+            }
 
             if (!TryLoadResources())
             {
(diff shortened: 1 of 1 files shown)


zeroZshadow:
Tested in standalone client, no issues

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -81,45 +84,44 @@ public IEnumerator CleanAndBuild(int worldLayers = 4, string seed = "SpatialOS G
...
-
-            mapTemplate.defaultTileCollection.LoadAndOptimizeTiles();
-            SceneManager.SetActiveScene(activeScene);
+                foreach (var tileCollection in mapTemplate.tileCollections)
+                {
+                    tileCollection.LoadAndOptimizeTiles();
+                }
 
-            yield return null;
+                mapTemplate.defaultTileCollection.LoadAndOptimizeTiles();
```
jessicafalk: why did that `yield return null;` go away

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -442,5 +445,31 @@ public static Vector3 GetWorldDimensions(int layers)
...
         }
+
+        private class SceneScope : IDisposable
+        {
+            private readonly Scene activeScene;
+
+            public SceneScope(Scene newActive)
+            {
+#if UNITY_EDITOR
+                if (Application.isPlaying)
```
jessicafalk: why do we need that? is `Application.isPlaying` only available in the editor?

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
zeroZshadow: This was an unneeded frame wait, removing it allowed me to simplify the code significantly

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
zeroZshadow: No, but there is no reason to do this test in standalone builds

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jessicafalk: is it also unneeded on mobile :D cause I don't want it to look like it's hanging

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
jessicafalk: but that code just doesn't look nice :(

workers/unity/Assets/Fps/Scripts/Config/MapBuilder.cs
```diff
@@ -281,36 +283,37 @@ private IEnumerator PlaceTiles()
 
             // Tiles are built in a spiral manner from the centre outward to ensure increasing the # of tile layers doesn't
             // alter the existing tile types.
-            SceneManager.SetActiveScene(tempScene);
-            for (var i = 0; i < tileCount; i++)
+            using (new SceneScope(tempScene))
```
jamiebrynes7: This does nothing

PR #200, opened 2019-06-12
Deferring the client-worker destroy by a frame

jessicafalk:
#### Description
have to defer destroying the worker by a frame to ensure everything gets cleaned up properly.

#### Tests
ran it, errors did not appear


workers/unity/Assets/Fps/Scripts/SetupLogic/ClientWorkerConnector.cs
```diff
@@ -41,6 +43,11 @@ public bool HasConnected()
             return Worker != null;
         }
 
+        public void DisconnectPlayer()
+        {
+            StartCoroutine(PrepareDestroy());
```
jamiebrynes7: We need to find a better pattern for this

Issue #212, opened 2019-07-15
Shooting a health pickup from the health pickups tutorial will throw an exception


jamiebrynes7:
**Affects:** Release v0.2.4 and up (may affect earlier versions as well)
**Internal Ticket:** [UTY-2130](https://improbableio.atlassian.net/browse/UTY-2130)

---

#### Description

After following the [health packs tutorial](https://docs.improbable.io/unity/alpha/projects/fps/tutorial), shooting the health pickups results in an stack trace like the following:

```
ArgumentException: A component with type:Component has not been added to the entity.
Unity.Entities.EntityComponentStore.AssertEntityHasComponent (Unity.Entities.Entity entity, Unity.Entities.ComponentType componentType) (at Library/PackageCache/com.unity.entities@0.0.12-preview.33/Unity.Entities/EntityComponentStore.cs:488)
Unity.Entities.EntityComponentStore.AssertEntityHasComponent (Unity.Entities.Entity entity, System.Int32 componentType) (at Library/PackageCache/com.unity.entities@0.0.12-preview.33/Unity.Entities/EntityComponentStore.cs:494)
Unity.Entities.ComponentDataFromEntity`1[T].get_Item (Unity.Entities.Entity entity) (at Library/PackageCache/com.unity.entities@0.0.12-preview.33/Unity.Entities/Iterators/ComponentDataFromEntity.cs:116)
Improbable.Gdk.Guns.ServerShootingSystem.OnUpdate () (at Packages/com.improbable.gdk.guns/Systems/ServerShootingSystem.cs:44)
```

#### Workaround

Don't shoot the health pickups. 😛

paulbalaji:
A quick-fix for this issue has been merged, and will be part of the next release.

jamiebrynes7:
Released in [v0.2.6](https://github.com/spatialos/gdk-for-unity-fps-starter-project/releases/tag/0.2.6)!

PR #220, opened 2019-07-23
Fix exception when shooting health pickups

paulbalaji:
#### Description

- check if gun data exists before using it
	- Quick-fix for #212 
- drive-by update to changelog

#### Tests

- shot some pickups and no more exceptions

#### Documentation

- [x] changelog

#### Primary reviewers


workers/unity/Packages/io.improbable.gdk.guns/Systems/ServerShootingSystem.cs
```diff
@@ -24,6 +23,11 @@ protected override void OnCreate()
         protected override void OnUpdate()
         {
             var events = componentUpdateSystem.GetEventsReceived<ShootingComponent.Shots.Event>();
+            if (events.Count == 0)
```
paulbalaji: this is not necessary - but it was annoying me

PR #226, opened 2019-08-16
FPS & `14.0.1` Worker SDK

jamiebrynes7:
#### Description
Upgraded the FPS project to the `14.0.1` Worker SDK.

- Update pinned
- Remove old arguments from worker JSONs
- Remove `public static Coordinates ToSpatialCoordinates(this Vector3 unityVector3)` since this method is available in transforms + we provide a static method in Coordinates for the same purpose.
- Connection flow changes

**Remaining work**
- [x] Changelog

PR #228, opened 2019-08-19
Fix Sim Player Coordinator

jamiebrynes7:
#### Description
Accidentally removed this one in #226. We actually need it to determine if we are a local/cloud deployment.


jamiebrynes7:
Cloud deployment was back to normal operation. Merging!

PR #238, opened 2019-11-05
Change network connection types 

paulbalaji:
#### Description

- client/simplayer KCP --> MUDP
- gamelogic/simplayercoordinator --> MUDP in editor, TCP otherwise

#### Tests

- tested locally with gamelogic in editor/built out

#### Documentation

- changelog

#### Primary reviewers


workers/unity/Assets/Fps/Scripts/SetupLogic/SimulatedPlayerCoordinatorWorkerConnector.cs
```diff
@@ -59,21 +60,26 @@ protected override async void Start()
...
-                builder.SetConnectionFlow(new ReceptionistFlow(workerId, new CommandLineConnectionFlowInitializer()));
+                connectionFlow = new ReceptionistFlow(workerId, new CommandLineConnectionFlowInitializer());
+                connectionParameters = CreateConnectionParameters(WorkerUtils.SimulatedPlayerCoordinator,
+                    new CommandLineConnectionParameterInitializer());
             }
 
-            return builder;
+            return new SpatialOSConnectionHandlerBuilder()
+                .SetConnectionFlow(connectionFlow)
+                .SetConnectionParameters(connectionParameters);
```
jamiebrynes7: This could be further refactored if you like: 


```
var workerId = CreateNewWorkerId(WorkerUtils.SimulatedPlayerCoordinator);
var initializer = Application.isEditor ? null : new CommandLineConnectionParameterInitializer();

return new SpatialOSConnectionHandlerBuilder()
	.SetConnectionFlow(new ReceptionistFlow(workerId, initializer)
	.SetConnectionParameters(CreateConnectionParameters(WorkerUtils.SimulatedPlayerCoordinator, initializer);

```

paulbalaji: big brain move - we don't need the isEditor check at all because the arguments would not be there anyway? the null would just save us the cost of attempting to get the values

jamiebrynes7: I do kind of prefer the explicit `isEditor` check. Otherwise its somewhat less clear

paulbalaji: in which case your code sample would actually have to be:

```csharp
var workerId = CreateNewWorkerId(WorkerUtils.SimulatedPlayerCoordinator);
var flowInitializer = Application.isEditor ? null: new CommandLineConnectionFlowInitializer();
var paramInitializer = Application.isEditor ? null : new CommandLineConnectionParameterInitializer();

return new SpatialOSConnectionHandlerBuilder()
	.SetConnectionFlow(new ReceptionistFlow(workerId, flowInitializer))
	.SetConnectionParameters(CreateConnectionParameters(WorkerUtils.SimulatedPlayerCoordinator, paramInitializer));
```

I prefer the current if (Application.isEditor) {} else {} block to that

jamiebrynes7: Sure 👍

PR #245, opened 2020-01-30
Add argument passthrough to init.sh

zeroZshadow:
#### Description
Add argument passthrough to init.sh and bootstrap.sh
Drive by - mark codegen as unsafe


ci/bootstrap.sh
```diff
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 set -e -u -o pipefail
 
+ARGS_PASSTHROUGH=$@
```
paulbalaji: inline again pls

PR #251, opened 2020-02-27
Add linting

jamiebrynes7:
Following on from https://github.com/spatialos/gdk-for-unity/pull/1306 this adds linting to the FPS Project


jamiebrynes7:
> This PR is the BOM

PR #271, opened 2020-05-19
Worker connector updates & refactoring

jamiebrynes7:
#### Description

Updated the `WorkerConnector` implementations with these https://github.com/spatialos/gdk-for-unity/pull/1365 changes. Additionally made all the map building use `Task` rather than coroutines and did some refactoring. 

#### Tests

- [x] Ran it locally
- [x] Release QA pipeline
- [ ] Test built out client

#### Documentation

- [x] Changelog

### DIFF
--- a/workers/unity/Assets/Fps/Scripts/WorkerConnectors/SimulatedPlayerCoordinatorWorkerConnector.cs
+++ b/workers/unity/Assets/Fps/Scripts/WorkerConnectors/SimulatedPlayerCoordinatorWorkerConnector.cs
@@ -30,7 +30,7 @@ namespace Fps.WorkerConnectors
         public int MaxSimulatedPlayerCount = 1;
         public int SimulatedPlayerCreationInterval = 5;
 
-        public Bounds Bounds { get; set; }
+        public Bounds Bounds { get; private set; }
 
         private readonly Dictionary<EntityId, List<GameObject>> proxies = new Dictionary<EntityId, List<GameObject>>();
         private readonly List<EntityId> localSimulatedPlayers = new List<EntityId>();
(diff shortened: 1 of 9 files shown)


workers/unity/Assets/Fps/Scripts/WorkerConnectors/SimulatedPlayerCoordinatorWorkerConnector.cs
```diff
@@ -30,6 +30,8 @@ public class SimulatedPlayerCoordinatorWorkerConnector : WorkerConnectorBase
         public int MaxSimulatedPlayerCount = 1;
         public int SimulatedPlayerCreationInterval = 5;
 
+        public Bounds Bounds { get; set; }
```
zeroZshadow, line 33: private set?

workers/unity/Assets/Fps/Scripts/WorkerConnectors/WorkerConnectorBase.cs
```diff
@@ -44,34 +27,42 @@ public override void Dispose()
...
-                mapTemplate,
-                worldSize,
-                transform,
-                Worker.WorkerType,
-                this);
+            LevelInstance = await MapBuilder.GenerateMap(mapTemplate, worldSize, transform, Worker.WorkerType);
         }
 
-        protected int GetWorldSize()
+        protected async Task WaitForWorkerFlags(CancellationToken token, params string[] flagKeys)
```
zeroZshadow, line 42: This looks like it is worth integrating into the core later?

workers/unity/Assets/Fps/Scripts/WorkerConnectors/SimulatedPlayerCoordinatorWorkerConnector.cs
jamiebrynes7: Ah yes - I meant to make it private set 😅

zeroZshadow:
Random through, does the map generator still work in the editor with this? Not when playing.

jamiebrynes7:
> Random through, does the map generator still work in the editor with this? Not when playing.

Yup - just tested with a map w/ 200 tile layers, Editor pauses for a few seconds and then completes fine

jamiebrynes7:
Everything looked good on the quick playtest! 📈

PR #285, opened 2020-07-21
Update Worker SDK Version to 14.7.0

seanjparker:
For [UTY-2662](https://improbableio.atlassian.net/browse/UTY-2662)

#### Description

- Updated the Worker SDK version to 14.7.0

#### Tests
Tests are green. Need to test:

- [x] Windows
- [x] Linux
- [ ] MacOS
- [ ] iOS
- [ ] Android

#### Documentation
How is this documented (for example: release note, upgrade guide, feature page, in-code documentation)?
 -  [ ] Add Changelog entry


spatialos.json
```diff
@@ -1,5 +1,5 @@
 {
     "name": "unity_gdk",
     "project_version": "0.0.1",
-    "sdk_version": "14.6.1"
```
paulbalaji, line 4: This doesn't actually change the worker SDK version used.

To test https://github.com/spatialos/gdk-for-unity/pull/1434, you'll have to update the `gdk.pinned` file to match a commit from that branch (so that you use the GDK with the worker sdk update), and then once the Core PR is merged update the `gdk.pinned` files in FPS/Blank with the commit hash from the `develop` branch.

paulbalaji: TL;DR - Without updating `gdk.pinned` you're using the GDK as of the 0.3.8 release, and not the version with your worker SDK update.

seanjparker: Oh right! I'll update `gdk.pinned` when the core PR is merged 👍

PR #298, opened 2021-01-21
Upgrade to v15 & latest develop

jamiebrynes7:
Most of this is component set changes, note the server interest changed since there's explicitly a single server which load balances, the interest on the LB partition covers everything in the deployment.


paulbalaji (approved):

workers/unity/Assets/Config/DeploymentLauncherConfig.asset
```diff
@@ -31,6 +31,7 @@ MonoBehaviour:
     SimulatedPlayerDeploymentConfigs:
     - Name: my_deployment_sim1
       SnapshotPath: 
+      SnapshotPath: snapshots/cloud.snapshot
```
zeroZshadow, line 33: Is github drunk, or is this property saved twice?

workers/unity/Assets/Fps/Scripts/Editor/SnapshotGenerator/SnapshotMenu.cs
```diff
@@ -25,6 +25,7 @@ private static void GenerateFpsSnapshot()
         private static Snapshot GenerateDefaultSnapshot()
         {
             var snapshot = new Snapshot();
+            snapshot.AddEntity(new EntityId(1), FpsEntityTemplates.CreateLoadBalancingPartition());
```
zeroZshadow: Is it required to be 1, or does any ID work?

zeroZshadow (approved):

workers/unity/Assets/Fps/Scripts/Editor/SnapshotGenerator/SnapshotMenu.cs
jamiebrynes7: It can be any entity ID, but this is referenced in the `BaseTemplate()` which is used for all entities. 

This ensures that the load balancing worker can claim this partition and then get authority over `AuthorityDelegation` for all entities.

workers/unity/Assets/Config/DeploymentLauncherConfig.asset
jamiebrynes7: 😕

workers/unity/Assets/Fps/Scripts/Editor/SnapshotGenerator/SnapshotMenu.cs
zeroZshadow: It's not very clear that it NEEDS to be 1, and is linked with the number in `BaseTemplate`

jamiebrynes7: Made this more explicit in https://github.com/spatialos/gdk-for-unity-fps-starter-project/pull/298/commits/05e3fe19d3e362a68674aa82f485c9eb7c03d672

workers/unity/Assets/Config/DeploymentLauncherConfig.asset
jamiebrynes7: I'm very confused, I don't even have this change in my local branch

jamiebrynes7: Managed to resolve it

